PR #13, opened 2024-11-14
INTPYTHON-415 Add _aget_relevant_documents to ParentDocumentRetriever

caseyclements:
Adds _aget_relevant_documents that runs sync in executor.


blink1073 (approved):
LGTM!

PR #23, opened 2024-12-05
INTPYTHON-445 Enable integration tests

blink1073:


blink1073:
This approach won't work because secrets can't be access from pull requests from forks.  Instead we should use local atlas.

PR #29, opened 2024-12-06
INTPYTHON-447 Finish integration tests

blink1073:
- Remove unused test files
- Remove whitespace changes in snapshot files
- Only run on one python version to avoid parallel write contention
- Do not fail-fast: let the builds complete so we can debug


caseyclements (approved):
lgtm

caseyclements (approved):
LGTM

PR #36, opened 2024-12-12
INTPYTHON-449 Use local atlas for testing

blink1073:


caseyclements (approved):
LGTM.  We can merge this to unblock the langgraph release.

PR #42, opened 2024-12-17
INTPYTHON-270  Implement a MongoDB record manager

blink1073:
Use `run_in_executor` uniformly for now so we don't need to depend on motor.  Once we finish async pymongo we can use that instead.


caseyclements (approved):
Looks good. Had a thorough review.

PR #43, opened 2024-12-17
INTPYTHON-455 Add CHANGELOG

blink1073:


caseyclements (changes requested):
I find the two packages inter-stitched to be hard to follow. Chronological order does not feel like the natural way to read this. I suggest that we choose either to place changelogs in the libs beside their respective pyproject.toml files or, break the single one into a section per package. Something like the following. Consider that it is likely that more packages will be added. 

# Changelog
--
## langchain-mongodb
###  0.3 (2024/12/13)
- Added ..

## langchain-checkpoint-mongodb
### 0.1 (2024/12/13)
- Added ..

## langwhizbang-mongodb
###  1.0 (2025/02/09)
- Added all singing all dancing robot

caseyclements (approved):
Much better! GTM

PR #44, opened 2024-12-17
INTPYTHON-459 Refactor handling of client into a pytest fixture

blink1073:


README.md
```diff
@@ -23,6 +23,8 @@ It contains the following packages.
     - `MongoDBAtlasParentDocumentRetriever` - new feature, docs coming soon!
 - Indexing
     - `MongoDBRecordManager` - new feature, docs coming soon!
+- Document loader
+    - `MongoDBLoader` - new feature, docs coming soon!
 - LangGraph persistence
```
caseyclements: This feels misleading. Should we use qualified names? I'm for breaking this into sections by package.

caseyclements (changes requested):
Bit of a misleading PR name for what's mainly an refactor of client to a fixture ;)  (which I totally agree with btw). Let's discuss the README.md

blink1073:
Ah, drat, I never added the new files.

README.md
blink1073: Yeah, let's do that.  I'll make this PR just about the refactor with a new ticket.  In the changelog PR we can break out the README.

caseyclements (approved):
lgtm

PR #52, opened 2025-01-08
INTPYTHON-467 Switch from Makefile to Justfile

blink1073:


caseyclements (approved):
just do_it. I'm onboard.

PR #64, opened 2025-01-22
Remove uv workspaces

blink1073:
The workspace concept doesn't really apply to our repo, since "Workspaces are not suited for cases in which members have conflicting requirements, or desire a separate virtual environment for each member".


caseyclements (approved):
LGTM.

PR #66, opened 2025-01-23
INTPYTHON-330 GraphRAG

caseyclements:
Adds MongoDBGraphStore
"""
    GraphRAG is a ChatModel that provides responses to semantic queries.
    As in Vector RAG, we augment the Chat Model's training data
    with relevant information that we collect from  documents.

    In Vector RAG, one uses an "Embedding" model that converts both
    the query, and the potentially relevant documents, into vectors,
    which can then be compared, and the most similar supplied to the
    Chat Model as context to the query.

    In Graph RAG, one uses an Entity-Extraction model that converts both
    the query, and the potentially relevant documents, into graphs. These are
    composed of nodes that are entities and edges that are relationships.
    The idea is that the graph can find connections between entities and
    hence answer questions that require more than one connection.

    It is also about finding common entities in documents,
    combining the attributes found and hence providing richer context than Vector RAG,
    especially in certain cases.

    When a document is extracted, each entity is represented by a single
    MongoDB Document, and relationships are defined in a nested field named
    'relationships' that contains list of targets, types, and attributes. This schema allows MongoDB's 
    $graphLookup to traverse all edges from an arbitrary number of starting nodes.

    When a query is made, the model extracts the entities and relationships from it,
    then traverses the graph starting from each of the entities found.
    The connected entities and relationships form the context
    that is included with the query to the Chat Model.

"""


libs/langchain-mongodb/langchain_mongodb/graphrag/prompts.py
```diff
@@ -0,0 +1,171 @@
...
+## Overview
+You are a meticulous analyst tasked with extracting information from unstructured text
+to build a knowledge graph in a structured json format of entities (nodes) and their relationships (edges).
+The graph will be stored in a MongoDB Collection and traversed using $graphLookup
+from starting points of entity nodes matching names found in a query, and follow their relationships.
+
+Use the following as guidelines.
+
+- Simplicity: The graph should have as few entities and relationship types as needed to convey the information in the input.
+- Consistency: Connections can only be made if entities and relationships use consistent naming
```
blink1073:
```suggestion
- Consistency: Connections can only be made if entities and relationships use consistent naming.
```

libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
```diff
@@ -0,0 +1,482 @@
+from __future__ import annotations
+
+import json
+import logging
+from copy import deepcopy
+from importlib.metadata import version
+from typing import Any, Dict, List, Optional, Union
+
+try:
```
blink1073: We need to put this behind an `if TYPE_CHECKING` to avoid a runtime dependency on `typing_extensions`.

libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
```diff
@@ -0,0 +1,482 @@
...
+"""Represents an Entity in the knowledge graph with specific schema. See .schema"""
+
+
+class MongoDBGraphStore:
+    """GraphRAG DataStore
+
+    GraphRAG is a ChatModel that provides responses to semantic queries
+    based on a Knowledge Graph that an LLM is used to create.
+    As in Vector RAG, we augment the Chat Model's training data
+    with relevant information that we collect from  documents.
```
blink1073:
```suggestion
    with relevant information that we collect from documents.
```

libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
```diff
@@ -0,0 +1,482 @@
...
+
+    In Vector RAG, one uses an "Embedding" model that converts both
+    the query, and the potentially relevant documents, into vectors,
+    which can then be compared, and the most similar supplied to the
+    Chat Model as context to the query.
+
+    In Graph RAG, one uses an "Entity-Extraction" model that converts
+    text into Entities and their relationships, a Knowledge Graph.
+    Comparison is done by Graph traversal, finding entities connected
+    to the query prompts. These are then supplied to the Chat Model  as context.
```
blink1073:
```suggestion
    to the query prompts. These are then supplied to the Chat Model as context.
```

libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
```diff
@@ -0,0 +1,482 @@
...
+    The main difference is that GraphRAG's output is typically in a structured format.
+
+    GraphRAG excels in finding links and common entities,
+    even if these come from different articles. It can combin information from
```
blink1073:
```suggestion
    even if these come from different articles. It can combine information from
```

libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
```diff
@@ -0,0 +1,482 @@
...
+    distinct source providing richer context than Vector RAG in certain cases.
```
blink1073:
```suggestion
    distinct sources providing richer context than Vector RAG in certain cases.
```

libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
```diff
@@ -0,0 +1,482 @@
...
+
+    Here are a few examples of so-called multi-hop questions where GraphRAG excels:
+    - What is the connection between ACME Corporation and GreenTech Ltd.?
+    - Who is leading the SolarGrid Initiative, and what is their role?
+    - Which organizations are participating in the SolarGrid Initiative?
+    - What is John Doe’s role in ACME’s renewable energy projects?
+    - Which company is headquartered in San Francisco and involved in the SolarGrid Initiative?
+
+    In Graph RAG, one uses an Entity-Extraction model that interprets
+    text documents that it is given extracting the query,
```
blink1073:
```suggestion
    text documents that it is given and extracting the query,
```

libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
```diff
@@ -0,0 +1,482 @@
...
+    When a query is made, the model extracts the entities in it,
+    then traverses the graph to find connections.
+    The closest entities and their relationships form the context
+    that is included with the query to the Chat Model.
+
+    Consider this example Query: "Does John Doe work at MongoDB?"
+    GraphRAG can answer this question even if the following two statements come
+    from completely different sources.
+    - "Jane Smith works with John Doe."
+    - "Jane Smiths works at MongoDB."
```
blink1073:
```suggestion
    - "Jane Smith works at MongoDB."
```

libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
```diff
@@ -0,0 +1,482 @@
...
+        allowed_entity_types: List[str] = None,
+        allowed_relationship_types: List[str] = None,
+        entity_examples: str = None,
+        entity_name_examples: str = None,
+        validate: bool = False,
+        validation_action: str = "warn",
+    ):
+        """
+        Args:
+            collection: Collection representing an Entity Graph
```
blink1073: Please use periods consistently

libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
```diff
@@ -0,0 +1,482 @@
...
+    ) -> List[BulkWriteResult]:
+        """Extract entities and upsert into the collection.
+
+        Each entity is represented by a single MongoDB Document.
+        Existing entities identified in documents will be updated.
+
+        Args:
+            documents: list of textual documents and associated metadata.
+        Returns:
+            List containing metadata on entities inserted and updated, one value for each input document
```
blink1073:
```suggestion
            List containing metadata on entities inserted and updated, one value for each input document.
```

libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
```diff
@@ -0,0 +1,482 @@
...
+        # Strip the ```json prefix and suffix
+        json_string = (
+            response.content.removeprefix("```json").removesuffix("```").strip()
+        )
+        return json.loads(json_string)
+
+    def find_entity_by_name(self, name: str) -> Optional[Entity]:
+        """Utility to get Entity dict from Knowledge Graph / Collection.
+        Args:
+            name: _id string to look for
```
blink1073:
```suggestion
            name: _id string to look for.
```

libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
```diff
@@ -0,0 +1,482 @@
...
+        Returns:
+            List of Entity dicts if any match name
```
blink1073:
```suggestion
            List of Entity dicts if any match name.
```

libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
```diff
@@ -0,0 +1,482 @@
...
+        self,
+        starting_entities: List[str],
+        max_depth: Optional[int] = None,
+    ) -> List[Entity]:
+        """Traverse Graph along relationship edges to find connected entities.
+
+        Args:
+            starting_entities: Traversal begins with documents whose _id fields match these strings.
+            max_depth: Recursion continues until no more matching documents are found,
+            or until the operation reaches a recursion depth specified by the maxDepth parameter
```
blink1073: does this need to be indented to render properly?

libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
```diff
@@ -0,0 +1,482 @@
...
+        query: str,
+        chat_model: BaseChatModel = None,
+        prompt: ChatPromptTemplate = None,
+    ) -> AIMessage:
+        """Responds to a query given information found in Knowledge Graph.
+
+        Args:
+            query: Prompt before it is augmented by Knowledge Graph.
+            chat_model: ChatBot. Defaults to entity_extraction_model.
+            prompt: Alternative Prompt Template. Defaults to prompts.rag_prompt
```
blink1073:
```suggestion
            prompt: Alternative Prompt Template. Defaults to prompts.rag_prompt.
```

blink1073 (changes requested):
I made a first pass, reviewing everything but the implementation details of the MongoDBGraphStore.  I mostly have cosmetic feedback.  Excellent work!

libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
caseyclements: Ok no problem. What's the best practice? How does one tell what must be placed behind `TYPE_CHECKING` and what doesn't?

blink1073: Anything that is only needed for typing *could* be put under TYPE_CHECKING.  Anything that would result in a new runtime dep *must* be put under TYPE_CHECKING.

libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
caseyclements: Yes. The indentation is necessary.

libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
caseyclements: Is this what's required?
```
if TYPE_CHECKING:
    Entity: TypeAlias = Dict[str, Any]
    """Represents an Entity in the knowledge graph with specific schema. See .schema"""
 ```

File: libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py [hunk 15]
```diff
@@ -0,0 +1,482 @@
...
+from langchain_core.prompts.chat import ChatPromptTemplate, SystemMessagePromptTemplate
+from langchain_core.runnables import RunnableSequence
+from pymongo import MongoClient, UpdateOne
+from pymongo.collection import Collection
+from pymongo.driver_info import DriverInfo
+from pymongo.results import BulkWriteResult
+
+from langchain_mongodb.graphrag import example_templates, prompts
+
+from .prompts import rag_prompt
```
blink1073, line 21: Can we please make all of these constants UPPERCASE?  That is the convention used by the [sql_toolkit](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/agent_toolkits/sql/prompt.py) and I find it easier to tell what is a constant and what is a user input.

libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
blink1073: I mean the line that continues after `max_depth`, I think it needs to be indented further

blink1073 (changes requested):

caseyclements (commented):
You right.

libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
caseyclements: I changed the string names to UPPERCase. By the time, they are imported here, they aren't constants, though. They've been wrapped. It's a good idea. Draws attention to itself.

blink1073 (approved):
LGTM!

PR #73, opened 2025-01-31
INTPYTHON-504 Add DriverInfo to MongoClients

caseyclements:
Consistently add DriverInfo to MongoClient calls so that we track metadata.

### DIFF
--- a/libs/langchain-mongodb/langchain_mongodb/chat_message_histories.py
+++ b/libs/langchain-mongodb/langchain_mongodb/chat_message_histories.py
@@ -1,5 +1,6 @@
 import json
 import logging
+from importlib.metadata import version
 from typing import Dict, List, Optional
 
 from langchain_core.chat_history import BaseChatMessageHistory
@@ -9,6 +10,7 @@ from langchain_core.messages import (
     messages_from_dict,
 )
 from pymongo import MongoClient, errors
+from pymongo.driver_info import DriverInfo
 
 logger = logging.getLogger(__name__)
 
@@ -112,7 +114,12 @@ class MongoDBChatMessageHistory(BaseChatMessageHistory):
             self.client = client
         elif connection_string:
             try:
-                self.client = MongoClient(connection_string)
+                self.client = MongoClient(
+                    connection_string,
+                    driver=DriverInfo(
+                        name="Langchain", version=version("langchain-mongodb")
+                    ),
+                )
             except errors.ConnectionFailure as error:
                 logger.error(error)
         else:
--- a/libs/langchain-mongodb/langchain_mongodb/loaders.py
+++ b/libs/langchain-mongodb/langchain_mongodb/loaders.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import logging
+from importlib.metadata import version
 from typing import Dict, List, Optional, Sequence
 
 from langchain_community.document_loaders.base import BaseLoader
@@ -9,6 +10,7 @@ from langchain_core.documents import Document
 from langchain_core.runnables.config import run_in_executor
 from pymongo import MongoClient
 from pymongo.collection import Collection
+from pymongo.driver_info import DriverInfo
 
 logger = logging.getLogger(__name__)
 
@@ -80,7 +82,10 @@ class MongoDBLoader(BaseLoader):
             include_db_collection_in_metadata (bool): Flag to include database and
             collection names in metadata.
         """
-        client = MongoClient(connection_string)
+        client = MongoClient(
+            connection_string,
+            driver=DriverInfo(name="Langchain", version=version("langchain-mongodb")),
+        )
         collection = client[db_name][collection_name]
         return MongoDBLoader(
             collection,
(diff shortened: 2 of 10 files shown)


libs/langchain-mongodb/langchain_mongodb/vectorstores.py
```diff
@@ -99,20 +99,11 @@ class MongoDBAtlasVectorSearch(VectorStore):
...
-            DB_NAME = "langchain_test_db"
-            COLLECTION_NAME = "langchain_test_vectorstores"
-            ATLAS_VECTOR_SEARCH_INDEX_NAME = "langchain-test-index-vectorstores"
-
-            MONGODB_COLLECTION = client[DB_NAME][COLLECTION_NAME]
-
-            vector_store = MongoDBAtlasVectorSearch(
-                collection=MONGODB_COLLECTION,
+            vector_store = MongoDBAtlasVectorSearch.from_connection_string(
+                connection_string="mongodb+srv://user:password@clusterip.mongodb.net/?w=majority&appName=Cluster0"
```
caseyclements: Is it safe, to make this easier on the eyes, to remove all of the metadata in the connection_string? So it's just 
`connection_string="mongodb+srv://user:password@clusterip.mongodb.net`?

libs/langchain-mongodb/langchain_mongodb/loaders.py
```diff
@@ -80,7 +82,10 @@ def from_connection_string(
             include_db_collection_in_metadata (bool): Flag to include database and
             collection names in metadata.
         """
-        client = MongoClient(connection_string)
+        client = MongoClient(
+            connection_string,
+            driver=DriverInfo(name="Langchain", version=version("langchain-mongodb")),
```
katcharov, line 87: We could use the official capitalization `LangChain`, or use `langchain-mongodb` if we need to differentiate from langgraph (though, can we just use `LangGraph`?). We should not need to report language specifically, because that will be known based on the driver.

I believe the version is intended to be the library version number, which should equal `0.2.69`. If it's not available, it seems the empty string is better.

caseyclements: I had similar questions originally. The decision to use Langchain, not LangChain was made long ago, and fees a number of BI (business intelligence) reports. `langchain-mongodb` is actually our package name. (e.g. `from langchain_mongodb.pipelines import vector_search_stage`). In any case, I am not starting from scratch here. I am just doing a review of our monorepo and adding the DriverInfo where it could have been added previously.

libs/langchain-mongodb/langchain_mongodb/vectorstores.py
blink1073: I think keeping the parts in that are part of the standard Atlas connection string makes sense

libs/langchain-mongodb/langchain_mongodb/vectorstores.py
```diff
@@ -99,20 +99,11 @@ class MongoDBAtlasVectorSearch(VectorStore):
...
+                namespace=f"db_name.collection_name",
```
blink1073: Unnecessary f-string

blink1073 (changes requested):

blink1073 (approved):
LGTM!

libs/langchain-mongodb/langchain_mongodb/chat_message_histories.py
```diff
@@ -112,7 +114,12 @@ def __init__(
             self.client = client
         elif connection_string:
             try:
-                self.client = MongoClient(connection_string)
+                self.client = MongoClient(
+                    connection_string,
+                    driver=DriverInfo(
+                        name="Langchain", version=version("langchain-mongodb")
```
prakul, line 120: Can we overload the version info to also include info about the specific feature? 
for instance here it would be version=version("langchain-mongodb-chat-message")

blink1073: Version is only tracked at the package level.  If we want to include more info, we usually add it to the name, like `PyMongo|c|async`, so in this case it would be `Langchain|chat-message`.

libs/langchain-mongodb/langchain_mongodb/vectorstores.py
prakul: Can we keep it to MONGODB_ATLAS_CLUSTER_URI

prakul (approved):

libs/langchain-mongodb/langchain_mongodb/loaders.py
katcharov: Alright, the main thing for the library naming is consistency.

Regarding the version, our handshake specification says [the following](https://github.com/mongodb/specifications/blob/a6897dc85129ddbb46ca540f4e10bca76aabae11/source/mongodb-handshake/handshake.md#supporting-wrapping-libraries):

> The version of the library wrapping the driver.

I would agree with what @blink1073 is [suggesting below](https://github.com/langchain-ai/langchain-mongodb/pull/73#discussion_r1943674585), which is adding the package or feature, if needed, to the name, for example `Langchain|chat-message` (and the versions would then be `0.2.69|`). This is what we would do in Java. Note that:

> The entire client metadata BSON document MUST NOT exceed 512 bytes.

This means that we have relatively few bytes to work with, and a long name like `langchain-mongodb-chat-message` increases the chance that the entire driver info is dropped from the message.

PR #83, opened 2025-02-20
Updates MongoDBGraphStore.__init__ signature

caseyclements:
Update signature of constructor to remove positional arguments, allows one of collection, or connection_string, database_name, and collection_name.

[Link to a passing patch](https://evergreen.mongodb.com/version/67b7501f4681210007ceeab7?redirect_spruce_users=true)

### DIFF
--- a/libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
+++ b/libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
@@ -130,18 +130,45 @@ class MongoDBGraphStore:
               - If "error", an exception will be raised if any document does not match the schema.
         """
         self._schema = deepcopy(entity_schema)
+        collection_existed = True
         if connection_string and collection is not None:
             raise ValueError(
                 "Pass one of: connection_string, database_name, and collection_name"
                 "OR a MongoDB Collection."
             )
-        if collection is not None:
+        if collection is None:  # collection is specified by uri and names
+            client: MongoClient = MongoClient(
+                connection_string,
+                driver=DriverInfo(
+                    name="Langchain", version=version("langchain-mongodb")
+                ),
+            )
+            db = client[database_name]
+            if collection_name not in db.list_collection_names():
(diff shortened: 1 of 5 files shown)


libs/langchain-mongodb/langchain_mongodb/graphrag/graph.py
```diff
@@ -122,6 +129,48 @@ def __init__(
...
+                    )
+                except OperationFailure:
+                    logger.warning(
+                        "Validation will NOT be performed. "
+                        "User must be DB Admin to add validation **after** a Collection is created. \n"
+                        "Please add validator when you create collection: "
+                        "db.create_collection.(coll_name, validator={'$jsonSchema': schema.entity_schema})"
+                    )
+        else:
+            client: MongoClient = MongoClient(
```
blink1073: We can't assume that the collection doesn't already exist just because we're creating the client.

caseyclements: @blink1073 This is inside a warning message.

blink1073: My question is: if the collection already exists, what is the behavior here?

blink1073: Actually, I see what happens now.  There would be no validation added.  I think we need to always attempt to add validation.

caseyclements: We do *attempt*. It's in a try/except block. Let's discuss on slack if we want to delay release of a version for this.

blink1073: We do not attempt for a collection that already exists, which may have been created by the user, no?

caseyclements:
Link to a [new passing patch](https://spruce.mongodb.com/version/67bde519397df30007bbd01b/) after additional cases of adding validation

PR #91, opened 2025-03-05
INTPYTHON-495 Automatically create indexes for checkpointer collections

caseyclements:
https://jira.mongodb.org/browse/INTPYTHON-495

Note that the index creation for `AsyncMongoDBSaver` is done in a `setup` method, following what is done in the Postgres checkpointer. This is done automatically if checkpointer is created with `from_conn_string`.

### DIFF
--- a/libs/langgraph-checkpoint-mongodb/langgraph/checkpoint/mongodb/aio.py
+++ b/libs/langgraph-checkpoint-mongodb/langgraph/checkpoint/mongodb/aio.py
@@ -115,7 +115,7 @@ class AsyncMongoDBSaver(BaseCheckpointSaver):
     ) -> AsyncIterator["AsyncMongoDBSaver"]:
         """Create asynchronous checkpointer
 
-        This includes creation of collections and indexes if they don't exist.
+        This includes creation of collections and indexes if they don't exist
         """
         client: Optional[AsyncIOMotorClient] = None
         try:
--- a/libs/langgraph-checkpoint-mongodb/langgraph/checkpoint/mongodb/aio.py
+++ b/libs/langgraph-checkpoint-mongodb/langgraph/checkpoint/mongodb/aio.py
@@ -84,6 +84,25 @@ class AsyncMongoDBSaver(BaseCheckpointSaver):
         self.writes_collection = self.db[writes_collection_name]
         self.loop = asyncio.get_running_loop()
 
+    async def setup(self):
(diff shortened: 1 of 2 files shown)


libs/langgraph-checkpoint-mongodb/langgraph/checkpoint/mongodb/aio.py
```diff
@@ -94,16 +113,23 @@ async def from_conn_string(
...
             client = AsyncIOMotorClient(conn_string)
-            yield AsyncMongoDBSaver(
+            saver = AsyncMongoDBSaver(
                 client,
                 db_name,
                 checkpoint_collection_name,
                 writes_collection_name,
                 **kwargs,
             )
+            await saver.setup()
```
vbarda: (nit) we usually recommend that users manually call `.setup()` on their checkpointers the first time

caseyclements: I saw that, in the postgres version. Why is that?

caseyclements: My preference is to allow users to use from_conn_string to automatically set up collections and indexes. Vadym and I discussed, and he doesn't have objection. One is free to set up collections and indexes manually, and/or call the constructor and setup separately. I'm off from this afternoon. Would love an approval on this.

blink1073 (approved):
LGTM!

libs/langgraph-checkpoint-mongodb/langgraph/checkpoint/mongodb/aio.py
```diff
@@ -82,8 +82,31 @@ def __init__(
...
+                unique=True,
+            )
+        if len(await self.writes_collection.list_indexes().to_list()) < 2:
+            await self.writes_collection.create_index(
+                keys=[
+                    ("thread_id", 1),
+                    ("checkpoint_ns", 1),
+                    ("checkpoint_id", -1),
+                    ("idx", 1),
+                ],
```
kirthiprakash, line 105: I'm a bit late, but I wanted to understand the reasoning behind the field order in this index.  

MongoDB recommends placing the field with the highest cardinality first, followed by lower-cardinality fields. After inspecting our data, I found that `checkpoint_id` has the highest cardinality. Is this the case globally, or is it specific to our dataset?

PR #92, opened 2025-03-12
INTPYTHON-386 Update Vectorstore constructors to check for search indexes and create if not found

blink1073:


caseyclements (approved):
LGTM

PR #95, opened 2025-03-17
INTPYTHON-434 Add standard test suites for Vector Store and Tools

blink1073:
Passing build: https://spruce.mongodb.com/version/67d95b1e6263880007a8ff22/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/libs/langchain-mongodb/langchain_mongodb/vectorstores.py
+++ b/libs/langchain-mongodb/langchain_mongodb/vectorstores.py
@@ -10,6 +10,7 @@ from typing import (
     Iterable,
     List,
     Optional,
+    Sequence,
     Tuple,
     TypeVar,
     Union,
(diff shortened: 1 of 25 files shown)


libs/langchain-mongodb/langchain_mongodb/vectorstores.py
```diff
@@ -396,24 +431,21 @@ def bulk_embed_and_insert_texts(
...
+                "_id": str_to_oid(i),
+                self._text_key: t,
+                self._embedding_key: embedding,
+                **m,
+            }
+            for i, t, m, embedding in zip(ids, texts, metadatas, embeddings)
+        ]
+        operations = [ReplaceOne({"_id": doc["_id"]}, doc, upsert=True) for doc in docs]
         # insert the documents in MongoDB Atlas
-        insert_result = self._collection.insert_many(to_insert)  # type: ignore
```
caseyclements, line 415: Does this change cause a performance hit? bulk_write(ReplaceOne) vs insert_many?

What was going here that the ids behaviour had to change?

libs/langchain-mongodb/tests/integration_tests/conftest.py
```diff
@@ -19,13 +21,8 @@ def technical_report_pages() -> List[Document]:
...
 @pytest.fixture
-def connection_string() -> str:
-    return os.environ["MONGODB_URI"]
-
-
-@pytest.fixture
-def client(connection_string: str) -> MongoClient:
-    return MongoClient(connection_string)
+def client() -> MongoClient:
+    return MongoClient(CONNECTION_STRING)
```
caseyclements, line 25: What's the new best-practice here? close the client?  Should we yield then close?

caseyclements (changes requested):
This looks good, although the changes coming from langchain-tests are opaque. Let's discuss.

A couple inline questions.

libs/langchain-mongodb/langchain_mongodb/vectorstores.py
blink1073: We need to use `ReplaceOne` and `upsert` to handle the case where a document is being overwritten.

libs/langchain-mongodb/tests/integration_tests/conftest.py
blink1073: We can't use fixtures because the tests are relying on properties.  So we're stuck with the warnings at shutdown.

libs/langchain-mongodb/langchain_mongodb/vectorstores.py
caseyclements: There wasn't a test for that case? Ouch.

PR #99, opened 2025-03-19
[VectorStores] Override _cosine_relevance_score_fn to use cosine similarity

Dheeraj-Nalapat:
## Description  
This PR fixes an issue in **MongoDBAtlasVectorSearch** where `_cosine_relevance_score_fn` incorrectly applies **1 - distance** normalization. Since MongoDB already returns **cosine similarity scores**, this normalization is **unnecessary**. The method is overridden to return the distance directly.    

## Dependencies  
No new dependencies introduced.

### DIFF
--- a/libs/langchain-mongodb/langchain_mongodb/vectorstores.py
+++ b/libs/langchain-mongodb/langchain_mongodb/vectorstores.py
@@ -261,6 +261,22 @@ class MongoDBAtlasVectorSearch(VectorStore):
     def collection(self, value: Collection) -> None:
         self._collection = value
 
+    @staticmethod
+    def _cosine_relevance_score_fn(distance: float) -> float:
+        """Return the raw cosine similarity score.
+
+        This method overrides the default behavior in `VectorStore`,
+        as MongoDB Atlas Vector Search provides scores directly in
+        cosine similarity format. No normalization is required.
+
+        Args:
+            distance (float): The cosine similarity score from MongoDB.
+
+        Returns:
+            float: The raw cosine similarity score.
+        """
+        return distance
+
     def _select_relevance_score_fn(self) -> Callable[[float], float]:
         scoring: dict[str, Callable] = {
             "euclidean": self._euclidean_relevance_score_fn,


caseyclements:
Hi @Dheeraj-Nalapat. Thank you for pointing this out! It is definitely misleading. In reality, we are not using `_select_relevance_score_fn`. Where the *argument* `relevance_score_fn` is used is to construct the vector search indexes. 

The base class `VectorStore` does not implement `_select_relevance_score_fn`, so it looks as though our team did, long ago. I will open a ticket to see what happens if we also do not implement  `_select_relevance_score_fn`.

Dheeraj-Nalapat:
Hi @caseyclements, thank you for taking the time to review this PR.

I wanted to clarify that `_select_relevance_score_fn` is not used when creating search indexes. Instead, the similarity metric is simply passed as a string to the relevant function from pymongo.

However, when using `MongoDBVectorStore` as a retriever via the `as_retriever` function, the base class `VectorStore` calls `_similarity_search_with_relevance_scores`, which in turn uses `_select_relevance_score_fn` to determine the relevance score logic.

Please let me know if you need any further details.
Again, I appreciate your time and feedback!

caseyclements:
Hi @Dheeraj-Nalapat. You're quite right! I missed that upon my search last night. According to our documentation [Atlas Vector Search Score](https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-stage/#atlas-vector-search-score) will always be normalized in [0,1], and so it seems like we should return distance unchanged not just for cosine, but in all cases. Do you agree?

caseyclements:
I made the following change of a branch and did some testing. Would you try it out in your environment? We're very pleased to have community contributions, so if you do the following, we'll definitely merge it.

```
    def _select_relevance_score_fn(self) -> Callable[[float], float]:
        """All Atlas Vector Search Scores are normalized in [0,1] so no change needed."""
        return lambda score: score
```

Dheeraj-Nalapat:
Hi @caseyclements, thanks for the update! I appreciate the clarification and agree that _select_relevance_score_fn should be replaced, as Atlas Vector Search always returns scores normalized in [0,1], making additional transformations unnecessary.

I’ve applied the same changes to my PR so that either can be merged without delays. Let me know if there's anything else you'd like me to modify.

Thanks again for your time and for guiding this improvement!

Dheeraj-Nalapat:
Hi @caseyclements, just following up on this. Let me know if there's anything else needed from my side to move this forward. Thanks again!

blink1073 (approved):
LGTM, thanks @Dheeraj-Nalapat!

PR #114, opened 2025-04-07
INTPYTHON-571 Add MongoDBDocStore to Readme

blink1073:
Also show the base class where appropriate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,32 +14,35 @@ It contains the following packages.
 
 ### LangChain
 
-- Vector store
+- Vector store (VectorStore)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,32 +14,35 @@ It contains the following packages.
 
 ### LangChain
 
-- Vector store
+- Vector store (VectorStore)
```
caseyclements, line 17: You're adding these base classes to disambiguate? I'm not sure if it begs more questions or answers.

blink1073: Perhaps if we said (subclass of \<Link to API docs>)?

caseyclements: Perhaps if we just had a new category?
- Key-Value Stores

caseyclements (approved):
LGTM. Lets go with what you've got for now.

Issue #130, opened 2025-05-08
Regex Aggregation Fails on Certain Strings in MongoDB Query


nahidosen:
We encountered an issue when using regex-based aggregation queries.

This query works correctly:

```python
'db.projects.aggregate([{ "$match": { "project_name": { "$regex": "Langchain", "$options": "i" } } }])'
```

But this query does **not** return results:

```python
'db.projects.aggregate([{ "$match": { "project_name": { "$regex": "Langchain Mongodb", "$options": "i" } } }])'
```

### **Expected Behavior:**

Both queries should return matching documents from the `projects` collection, applying the regex case-insensitively.

### **Observed Behavior:**

* Simple patterns (like `"Langchain"`) work.
* Complex patterns or strings with spaces (like `"Langchain Mongodb"`) fail, even though the data exists.

### **Environment:**

Python version: 3.11
```bash
langchain                    0.3.25
langchain-core               0.3.59
langchain-mongodb            0.6.1
langchain-openai             0.3.16
langchain-text-splitters     0.3.8
langgraph                    0.4.3
langgraph-checkpoint         2.0.25
langgraph-checkpoint-mongodb 0.1.3
pymongo                      4.11.3
```

### **Steps to Reproduce:**

1. Set up a `projects` collection with documents containing names like `"Langchain"` and `"Langchain Mongodb"`.
2. Run aggregation queries using regex through `langchain-mongodb`.
3. Observe that only simpler patterns return results.

### **Additional Notes:**

* The query works correctly when run directly in the MongoDB shell.
* The issue may stem from how LangChain passes regex patterns into the MongoDB driver or how it serializes the command.

Could you investigate if regex handling in LangChain MongoDB has limitations or bugs when the pattern includes spaces or more complex strings?

mhasan502:
This PR solves your issue #131

nahidosen:
thank you @mhasan502 , my issue is fixed #131

PR #131, opened 2025-05-08
Fix: Enhance _parse_command to robustly handle whitespace and formatting

mhasan502:
#130
Refined the `_parse_command` method from `langchain_mongodb/agent_toolkit/database.py` to normalize whitespace using regular expressions. This replaces the previous method of removing all spaces and newlines. This adjustment makes JSON parsing more resilient to different input formats, especially those resembling JavaScript syntax with irregular spacing.


blink1073 (approved):
Thank you!

blink1073:
We'll make a patch release with this fix early next week.

PR #138, opened 2025-05-21
Async Variante for MongoDBAtlasHybridSearchRetriever

DRXD1000:
The MongoDBAtlasHybridSearchRetriever is one of the best retrievers i have used so far. Since MongoDB now also supports an AsyncMongoClient i would like to properly use the retriever with ainvoke. 

I have done Unit Tests as well as Integration Tests (with local atlas and ollama) spelling and pre-commit checks.


DRXD1000:
@blink1073 would you mind looking at the implementation? Also i have not done typing so far. Help and feedback would be very much appreciated.

blink1073:
Hi @DRXD1000, we are trying to avoid having duplicate forms of the same logic.  We could instead copy there pattern we use in the ParentDocRetriver [here](https://github.com/langchain-ai/langchain-mongodb/blob/59a08e8cb4ae3592acb1ef7d4af3921aebcbf45f/libs/langchain-mongodb/langchain_mongodb/retrievers/parent_document.py#L124).

PR #139, opened 2025-05-22
Async Variante for MongoDBAtlasHybridSearchRetriever

DRXD1000:
Improved version of #138 .

All tests and checks done.


blink1073:
Thanks @DRXD1000, my position is that we should only accept synchronous objects in our classes, and then use `run_in_executor` in async methods.  That is how we've been avoiding duplicating logic.

DRXD1000:
Oh then this was a missunderstanding.

DRXD1000:
> Apologies for the confusion!

All good :)

PR #144, opened 2025-05-28
INTPYTHON-630 Fixes handling of ttlMonitorSleepSecs

caseyclements:
ttlMonitorSleepSecs was first reduced  programatically via `client.admin.command(setParameter,ttlMonitorSleepSecs)`. This reduced the time required to test the TTL feature from 60 seconds to 2. This is because 60s is the frequency that the background process is kicked off at.  However, this only worked on local Atlas. For Cloud Atlas, this feature requires that one  setParameter manually vi the Atlas Administration UI. This PR is small. It just wraps  client.admin.command(setParameter,ttlMonitorSleepSecs) in a try block. Crucially though, it comes at the same time as @blink1073 updated the parameter via API for our remote tests' cluster.

See docs [here](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Clusters/operation/updateClusterAdvancedConfiguration).


blink1073 (approved):
LGTM!

PR #148, opened 2025-06-09
INTPYTHON-629 Take 2 - Encouraged LLM to find as many entities and relationships as it could

caseyclements:
As LLM requests would often return without any results, I modified the prompts to look for 'potential entities' and find as many as it could. I also removed a number of guiding principles that contained which also would restrict an LLM, that used words like only this and unless that.


caseyclements:
Patch builds: 
* [here](https://spruce.mongodb.com/version/6846eb8b01104b0007bd6818/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
* [here](https://spruce.mongodb.com/version/6844a182fdc2f6000796e5ad/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
* and [here](https://spruce.mongodb.com/version/6846ea8bba9aad0007bf6da8/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) where in this last one, the remote failure is unrelated, having to do with an index failing to build in 2 minutes.

PR #150, opened 2025-06-13
INTPYTHON-638 Updated version and CHANGELOG for 0.1.4

caseyclements:


NoahStapp (commented):
Are the `uv.lock` changes here intended to be committed?

caseyclements:
> Are the `uv.lock` changes here intended to be committed?

Yes. They are in the pre-commit CI, so must be. There's no actual changes in pyproject. I was a bit surprised myself. It now just adds upload_time.

NoahStapp (approved):

Issue #165, opened 2025-07-16
Support Multi Field $search?


DRXD1000:
MongoDB supports multi-field $search operations: [Path-Construction](https://www.mongodb.com/docs/atlas/atlas-search/path-construction/?tck=mongodb_ai_chatbot)

Would this be a feature that is interesting for the langchain-mongodb package ? 

I have a Version available in my [fork](https://github.com/DRXD1000/langchain-mongodb/tree/mu/multi_field_text_search).

I wanted to clarify with you before i open a pull request.

PR #166, opened 2025-07-17
Add multi_field support for $search operation

DRXD1000:
Linked to #165 

All tests checked.


DRXD1000:
@blink1073  The current draft would expect to parse a list of strings -> list of fields to be indexed to the text_key parameter. All other process steps would expect the first key, to be the 'primary key' to read into the page_content field of the Langchain document.

What is your opinion on that. Would you prefer a class with like a 'primary' key ?

blink1073 (approved):
Great, thank you!  We'll make a release with this next week.

Issue #84, opened 2025-02-25
Weighted RRF


Shanorino:
In `MongoDBAtlasHybridSearchRetriever`, currently it's not possible to add weights separately to 2 retrievers (in other words, it's always 50-50).

A workaround is to change `vector_penalty` and `fulltext_penalty`. However, tuning the RRF parameter `k` changes the shape of the reciprocal rank function, while multiplying by a weight is a simple linear scaling of the entire curve. These two approaches are not mathematically equivalent and lead to different effects on how each rank contributes to the final score.

One thing to mention is that `EnsembleRetriever` in Langchain also implemented the weights (instead of changing the penalty constant `k`).

Thus, I'd propose implementing weights to the class `MongoDBAtlasHybridSearchRetriever` by adding 2 more stages to the Mongo pipeline, so that it can be instantiated this way:
```Python
retriever = MongoDBAtlasHybridSearchRetriever(
vectorstore = vector_store,
search_index_name = "search_index",
top_k = 5,
fulltext_weight=0.3,
vector_weight=0.7,
fulltext_penalty = 50,
vector_penalty = 50
)
```

blink1073:
Hi @Shanorino, I opened https://jira.mongodb.org/browse/INTPYTHON-542 to track this issue.

caseyclements:
Hi @Shanorino. Would you mind going into further detail on what changes you're suggesting? Maybe you can write down mathematically, or include a link?

s-xu-wafios:
Hi @caseyclements ,  so if the original RRF can be represented this way:

```math
\text{RRF}(d) = \sum_{i=1}^{n} \frac{1}{k + r_i(d)}
```

The weighted RRF would be like this:

```math
\text{WeightedRRF}(d) = \sum_{i=1}^{n} w_i \cdot \frac{1}{k + r_i(d)}
```

That said, I added a "weighted score stage", where each retriever has a `$multiply` with `weight`
```python
    def _weighted_score_stage(self, score_field: str, weight: float, penalty: float) -> List[Dict[str, Any]]:
        """
        Apply weighted scoring with custom penalty for either vector or full-text scores.
        """
        return [
            {"$group": {"_id": None, "docs": {"$push": "$$ROOT"}}},
            {"$unwind": {"path": "$docs", "includeArrayIndex": "rank"}},
            {
                "$addFields": {
                    f"docs.{score_field}": {
                        "$multiply": [
                            weight,
[...]

PR #168, opened 2025-07-25
feat(INTPYTHON-542): support weighted RRF

s-xu-wafios:
I added support fort weighted RRF (mentioned in issue https://github.com/langchain-ai/langchain-mongodb/issues/84)

### DIFF
--- a/libs/langchain-mongodb/langchain_mongodb/retrievers/hybrid_search.py
+++ b/libs/langchain-mongodb/langchain_mongodb/retrievers/hybrid_search.py
@@ -43,6 +43,10 @@ class MongoDBAtlasHybridSearchRetriever(BaseRetriever):
     """Penalty applied to vector search results in RRF: scores=1/(rank + penalty)"""
     fulltext_penalty: float = 60.0
     """Penalty applied to full-text search results in RRF: scores=1/(rank + penalty)"""
+    vector_weight: float = 0.5
+    """Weight applied to vector search results in RRF: score = weight * (1 / (rank + penalty + 1))"""
+    fulltext_weight: float = 0.5
+    """Weight applied to full-text search results in RRF: score = weight * (1 / (rank + penalty + 1))"""
     show_embeddings: float = False
     """If true, returned Document metadata will include vectors."""
     top_k: Annotated[
@@ -95,7 +99,7 @@ class MongoDBAtlasHybridSearchRetriever(BaseRetriever):
                 oversampling_factor=self.oversampling_factor,
             )
         ]
-        vector_pipeline += reciprocal_rank_stage("vector_score", self.vector_penalty)
+        vector_pipeline += reciprocal_rank_stage(score_field="vector_score", penalty=self.vector_penalty, weight=self.vector_weight)
 
         combine_pipelines(pipeline, vector_pipeline, self.collection.name)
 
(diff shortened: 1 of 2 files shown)


caseyclements:
[Passing patch of existing tests](https://spruce.mongodb.com/version/689b0bcc7965c700079e22ef)

s-xu-wafios:
Hi @caseyclements, I see that there are some failing tests. Not sure what was causing the issue. Do I need to do anything here?

libs/langchain-mongodb/langchain_mongodb/retrievers/hybrid_search.py
```diff
@@ -43,6 +43,10 @@ class MongoDBAtlasHybridSearchRetriever(BaseRetriever):
     """Penalty applied to vector search results in RRF: scores=1/(rank + penalty)"""
     fulltext_penalty: float = 60.0
     """Penalty applied to full-text search results in RRF: scores=1/(rank + penalty)"""
+    vector_weight: float = 0.5
```
caseyclements: @s-xu-wafios I did not see this earlier, but should the defaults not be 1.0  so that this is not a backwards breaking change (i.e. will change scores if one upgrades)? As the scores themselves have no physical meaning, serving only to rank documents, the weights need not sum to one. Am I correct?

caseyclements (dismissed):
Please respond to comment about default weights.

libs/langchain-mongodb/langchain_mongodb/retrievers/hybrid_search.py
s-xu-wafios: @caseyclements  You are right that the scores have no physical meanings, because we only care about the ranking in the end. But I would say both ideas are valid. Because
- on the one hand, for backwards compatibility, the old behavior was `(1 * vector_score) + (1 * text_score)`, so the defaulted weight `w_i` is `1`
- however, on the other hand, in `langchain` implementation, Class `EnsembleRetriever` (imported from `langchain.retrievers`) implemented it in a way (see method set_weights(), it has defaulted the weight to `1 / n`, not `1`): 
(source code in langchain project)
https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/retrievers/ensemble.py#L84
```python
class EnsembleRetriever(BaseRetriever):
    """Retriever that ensembles the multiple retrievers.

    It uses a rank fusion.

    Args:
        retrievers: A list of retrievers to ensemble.
        weights: A list of weights corresponding to the retrievers. Defaults to equal
            weighting for all retrievers.
        c: A constant added to the rank, controlling the balance between the importance
            of high-ranked items and the consideration given to lower-ranked items.
            Default is 60.
        id_key: The key in the document's metadata used to determine unique documents.
            If not specified, page_content is used.
    """

    @model_validator(mode="before")
    @classmethod
    def set_weights(cls, values: dict[str, Any]) -> Any:
        if not values.get("weights"):
            n_retrievers = len(values["retrievers"])
            values["weights"] = [1 / n_retrievers] * n_retrievers
        return values
```



In my opinion, `w_i = 1 / n` looks more "normalized" to me and it aligns better with the default behavior in langchain main library as well. 
But I think the decision is up to you. I can also gladly change the default weight to 1.

caseyclements: I agree that it's natural for weights generally to sum up to one, but in our case, backwards compatibility overrules. Thank you for this. I should have spotted it earlier.

s-xu-wafios: No problem, I have adjusted the default weights to `1.0` and pushed a new commit

PR #171, opened 2025-07-25
INTPYTHON-680 Use random seed when testing ChatOpenAI

blink1073:
Patch build: https://spruce.mongodb.com/version/6883f6f577ce2200074c3838/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC


caseyclements (approved):
How were things passing with the wrong dimensions before this? When did we switch to 4o without changing dims?

blink1073:
I had tried using a larger model to see if it would improve the accuracy of the test, but realized I'd also have to update the index definitions in the pipeline, so I used flaky instead.

PR #182, opened 2025-08-05
INTPYTHON-655 Infer dimensions from embedding if not provided

blink1073:
Patch build: https://spruce.mongodb.com/version/689271e34110f50007a71967/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Preserves previous behavior by not auto-creating unless either dimensions or auto_create_index are explicitly set.

### DIFF
--- a/libs/langchain-mongodb/tests/unit_tests/test_vectorstores.py
+++ b/libs/langchain-mongodb/tests/unit_tests/test_vectorstores.py
@@ -223,4 +223,4 @@ class TestMongoDBAtlasVectorSearch:
             collection=collection,
             index_name=INDEX_NAME,
         )
-        assert len(collection._search_indexes)
+        assert len(collection._search_indexes) == 0


libs/langchain-mongodb/langchain_mongodb/vectorstores.py
```diff
@@ -208,7 +208,7 @@ def __init__(
         embedding_key: str = "embedding",
         relevance_score_fn: str = "cosine",
         dimensions: int = -1,
-        auto_create_index: bool = True,
+        auto_create_index: bool | None = None,
```
Jibola, line 211: Why does this now default to `None` instead of `True`?

blink1073: Because the old behavior of "true" was actually only true if `dimensions` was set, so I had to tweak this value.  I wanted to preserve the behavior of "don't create an index unless the user explicitly asks for it".

PR #183, opened 2025-08-06
INTPYTHON-461 - Add additional client metadata on LangChain Integrations

NoahStapp:


libs/langchain-mongodb/langchain_mongodb/utils.py
```diff
@@ -20,14 +20,25 @@
...
 
 logger = logging.getLogger(__name__)
 
 Matrix = Union[List[List[float]], List[np.ndarray], np.ndarray]
 
+DRIVER_METADATA = DriverInfo(name="Langchain", version=version("langchain-mongodb"))
+
+
+def append_client_metadata(client: MongoClient) -> None:
+    # append_metadata was added in PyMongo 4.14.0, but is a valid database name on earlier versions
```
caseyclements, line 38: Should be a docstring. This is a note that explains why we use callable, but I suggest something like "Add simple driver and integration  information for server telemetry."

NoahStapp: This method should be internal, since we don't want end users using it themselves. Does a docstring still make sense in that context? The focus in my mind of the comment is communicating why we're using `callable`.

caseyclements: Sure. If its got a leading `_`, it should be fine.

caseyclements (changes requested):

caseyclements (approved):
LGTM pending `just` specifier issue spoiling the tests.

blink1073:
Hmm it seems to have been a regression, since it is happening in #166 as well.  We must not have caught it because of our filtering on when tests should run.

PR #204, opened 2025-09-11
Prototype with new class AutoEmbeddingVectorStore and test_autoembedd…

caseyclements:
This is just a draft PR, for testing and discussion.

### DIFF
--- a/libs/langchain-mongodb/langchain_mongodb/vectorstores.py
+++ b/libs/langchain-mongodb/langchain_mongodb/vectorstores.py
@@ -48,6 +48,10 @@ logger = logging.getLogger(__name__)
 DEFAULT_INSERT_BATCH_SIZE = 100
 
 
+# TODO: fold the autoembedding indexes into MongoDBAtlasVectorSearch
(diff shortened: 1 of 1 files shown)


libs/langchain-mongodb/langchain_mongodb/vectorstores.py
```diff
@@ -865,3 +867,285 @@ def create_vector_search_index(
...
+            docs.append(
+                (Document(page_content=text, metadata=res, id=res["_id"]), score)
+            )
+        return docs
+
+    @classmethod
+    def from_texts(
+        cls,
+        texts: List[str],
+        embedding: Embeddings = VoyageAIEmbeddings(model="voyage-3-large"),
```
prakul: This is not needed since the search index is handling this

caseyclements: This is a red herring. I added from_texts as it is abstractmethod. I wasn't using it for the test. `embedding ` is a required argument to the base `VectorStore`'s constructors.

PR #220, opened 2025-10-03
INPYTHON-764 Make usage of text_key more obvious

blink1073:
Patch build: https://spruce.mongodb.com/version/68e3e465cef4a80007fa957e/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Turning on warnings exposed some other places where we weren't handling warnings well, especially with `top_k`, which was always triggering a warning just by us accessing it.  There were also several places where we were not explicitly closing clients in tests.


caseyclements (approved):
This is great! Thank you. LGTM

caseyclements (changes requested):
Would you please link to a passing patch?

PR #208, opened 2025-09-18
INTPYTHON-837 Integrate pymongo-search-utils

aclark4life:
Replace functions in this library with identical functions from pymongo-search-utils

- [x] append_client_metadata
- [x] bulk_embed_and_insert_texts
- [x] combine_pipelines
- [x] create_fulltext_search_index
- [x] create_vector_search_index
- [x] drop_vector_search_index
- [x] final_hybrid_stage
- [x] reciprocal_rank_stage
- [x] text_search_stage
- [x] update_vector_search_index
- [x] vector_search_stage

Patch build: https://spruce.mongodb.com/version/6924b537f767320007b1a18d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #241, opened 2025-10-30
INTPYTHON-816 - Bump minimum Python version to 3.10

NoahStapp:
[INTPYTHON-816](https://jira.mongodb.org/browse/INTPYTHON-816)

## Summary

LangChain 1.0 requires Python 3.10+. Updating our own minimum version now before the package split will reduce the complexity of doing so.

## Changes in this PR

Bump minimum Python version across the board to 3.10. Update ruff to the latest release for fixes such as the removal of [UP038](https://docs.astral.sh/ruff/rules/non-pep604-isinstance/).

### DIFF
--- a/libs/langchain-mongodb/CHANGELOG.md
+++ b/libs/langchain-mongodb/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ---
 
+## Changes in version 0.7.2 (2025/10/30)
+
+- Pin LangChain version to < 1.0 to avoid compatibility issues while in pre-1.0.
+
 ## Changes in version 0.7.1 (2025/10/13)
 
 - Only list authorized collections when listing collections.
--- a/libs/langchain-mongodb/pyproject.toml
+++ b/libs/langchain-mongodb/pyproject.toml
@@ -5,7 +5,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "langchain-mongodb"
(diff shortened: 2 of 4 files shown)


libs/langchain-mongodb/langchain_mongodb/agent_toolkit/database.py
```diff
@@ -41,9 +41,9 @@ def __init__(
         self,
         client: MongoClient,
         database: str,
-        schema: Optional[str] = None,
-        ignore_collections: Optional[List[str]] = None,
-        include_collections: Optional[List[str]] = None,
+        schema: str | None = None,
+        ignore_collections: List[str] | None = None,
```
blink1073: I don't think we should enable this upgrade rule, it is causing a lot of unnecessary churn and likely to conflict with #208

NoahStapp: Sure we can turn it off.

NoahStapp: Turned off.

blink1073:
Failures are tracked in [INTPYTHON-799](https://jira.mongodb.org/browse/INTPYTHON-799)

blink1073 (approved):
LGTM!

PR #271, opened 2025-11-17
INTPYTHON-828 - Make Langchain-MongoDB compatible with LangChain >= 1.0

NoahStapp:
[INTPYTHON-828](https://jira.mongodb.org/browse/INTPYTHON-828)

Patch build (updated): https://spruce.mongodb.com/version/691cbf67fd4ec10007a99531/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/libs/langchain-mongodb/langchain_mongodb/agent_toolkit/database.py
+++ b/libs/langchain-mongodb/langchain_mongodb/agent_toolkit/database.py
@@ -341,10 +341,15 @@ class MongoDBDatabase:
                 raise ValueError("ObjectId must contain a value.")
             return f"ObjectId('{oid_str}')"
 
+        def _handle_id_key(match: Any) -> str:
+            return f'"{match.group(1)}"'
+
         patterns = [
             (r'ISODate\(\s*["\']([^"\']*)["\']\s*\)', _handle_iso_date),
             (r'new\s+Date\(\s*["\']([^"\']*)["\']\s*\)', _handle_new_date),
             (r'ObjectId\(\s*["\']([^"\']*)["\']\s*\)', _handle_object_id),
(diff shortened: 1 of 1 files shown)


libs/langchain-mongodb/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ---
 
+## Changes in version 0.8.0 (2025/11/18)
+
+- Update dependencies to align with LangChain 1.0, which is now the minimum supported version of LangChain.
```
blink1073, line 7: I think we should call out that we depend on `langchain_classic` now, since it is a heavy dependency that might not be obvious why it ended up in a user's environment.

caseyclements: > I think we should call out that we depend on `langchain_classic` now, since it is a heavy dependency that might not be obvious why it ended up in a user's environment.

These same dependencies (or close to) already exist in langchain < 1.0. What I would mention is that. "Note: langchain_mongodb 0.8.0 depends on langchain_classic, so still has the heavy dependency set than langchain did before the v1 refactor."

NoahStapp: Is the updated wording sufficient?

caseyclements: Did you call out langchain-classic?

NoahStapp: The latest commit does.

NoahStapp:
Looking into the test failure for `test_toolkit_response` with error `ValueError: Failed to parse aggregation pipeline: name '_id' is not defined`.

PR #298, opened 2026-01-05
INTPYTHON-865 Prepare for langchain-mongodb 0.10.0

aclark4life:
INTPYTHON-865

### Checklist for Author

### Checklist for Reviewer

### DIFF
--- a/libs/langchain-mongodb/uv.lock
+++ b/libs/langchain-mongodb/uv.lock
@@ -1174,7 +1174,7 @@ wheels = [
 
 [[package]]
 name = "langchain-mongodb"
-version = "0.9.0"
+version = "0.10.0"
 source = { editable = "." }
 dependencies = [
     { name = "langchain" },
--- a/uv.lock
+++ b/uv.lock
@@ -858,7 +858,7 @@ wheels = [
 
 [[package]]
 name = "langchain-mongodb"
-version = "0.9.0"
+version = "0.10.0"
 source = { editable = "libs/langchain-mongodb" }
 dependencies = [
     { name = "langchain" },


libs/langchain-mongodb/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ---
 
+## Changes in version 0.10.0 (2026/01/05)
+
+- Bump langchain-core dependency for CVE-2025-68664.
```
NoahStapp: We should have a link to the CVE.

NoahStapp (approved):

aclark4life:
> Lockfile needs to be updated.

I saw that, but I can't tell from the commits what that means. In 0.9.0 only the top level revision changed. In langgraph-checkpoint-mongodb 0.2 both parent and child uv.lock versions were updated. If I manually update the top level revision from 3 -> 4 then the pre-commit hook updates the child uv.lock. Maybe we should update the README with some instructions e.g. 
- Run `uv lock` in the root of the checkout.

?

NoahStapp:
> > Lockfile needs to be updated.
> 
> I saw that, but I can't tell from the commits what that means. In 0.9.0 only the top level revision changed. In langgraph-checkpoint-mongodb 0.2 both parent and child uv.lock versions were updated. If I manually update the top level revision from 3 -> 4 then the pre-commit hook updates the child uv.lock. Maybe we should update the README with some instructions e.g.
> 
>     * Run `uv lock` in the root of the checkout.
> 
> 
> ?

`uv sync` is the command you want to run here. That'll update the environment represented in the lockfile to be consistent with the toml changes.

PR #365, opened 2026-04-09
fix: handle authorizedCollections=True for Firestore MongoDB compatibility

jjjojoj:
## Summary

Firestore and other non-native MongoDB backends do not support the `authorizedCollections` parameter in `list_collection_names()`, causing an `OperationFailure` at initialization.

This wraps all three call sites (`database.py`, `cache.py`, `graphrag/graph.py`) in try/except so they gracefully degrade to the parameter-less version when a `PyMongoError` is raised.

Fixes langchain-ai/langchain#36609

## Changes

- **`agent_toolkit/database.py`**: wrap `list_collection_names(authorizedCollections=True)` in try/except
- **`cache.py`**: same fix for MongoDBCache initialization
- **`graphrag/graph.py`**: same fix for graph collection initialization

## Testing

Existing tests should pass. The try/except path only activates when `authorizedCollections` is rejected by the server, which is the exact scenario described in the issue.

### DIFF
--- a/libs/langchain-mongodb/langchain_mongodb/agent_toolkit/database.py
+++ b/libs/langchain-mongodb/langchain_mongodb/agent_toolkit/database.py
@@ -57,9 +57,12 @@ class MongoDBDatabase:
             )
         self._include_colls = set(include_collections or [])
         self._ignore_colls = set(ignore_collections or [])
-        self._all_colls = set(
-            self._db.list_collection_names(authorizedCollections=True)
-        )
+        try:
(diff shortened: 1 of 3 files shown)


NoahStapp:
Hello,

Google's official compatibility policy for FireStore and MongoDB is listed here: https://docs.cloud.google.com/firestore/mongodb-compatibility/docs/overview.

MongoDB doesn't offer support for Google FireStore. For help with this issue, contact [Firebase support](https://firebase.google.com/support).

PR #375, opened 2026-04-21
INTPYTHON-948 Fix multikey index collision on namespace array 

aclark4life:
INTPYTHON-948

## Summary

Replace the unique index on the namespace array + key with one on a denormalized namespace_str field (e.g. \`"users/alice/preferences"\`) + key. Indexing an array creates a multikey index whose entries are individual elements, so two documents sharing any namespace component and the same key would collide with a duplicate-key error.

On init, legacy documents are backfilled with \`namespace_str\` in batches, then the new \`(namespace_str, key)\` unique index is created, and finally the old \`(namespace, key)\` index is dropped — ensuring uniqueness is always enforced. All read and write operations (\`get\`, \`delete\`, \`batch\`) now filter on \`namespace_str\` + \`key\` to use the new index. Namespace parts are validated to not contain the separator to prevent join collisions.

## Changes in this PR

- Replaces multikey index on \`namespace\` array with a unique index on \`namespace_str\` + \`key\`
- Backfills \`namespace_str\` on legacy documents in batches of 1000 before building the unique index
- Drops the old \`(namespace, key)\` index only after the new unique index exists (no window without uniqueness enforcement)
- Migrates all query filters in \`get()\`, \`delete()\`, and \`batch()\` to use \`namespace_str\` + \`key\`
- Stores \`namespace\` array via \`\$setOnInsert\` on upsert so \`Item\` deserialization continues to work
- Raises \`ValueError\` if any namespace part contains the separator to prevent join collisions

## Test Plan

Integration tests cover the duplicate-key regression (INTPYTHON-948).

### Checklist for Author

### Checklist for Reviewer

- [x] Do you fully understand the implementation? (Would you be comfortable explaining how this code works to someone else?)

### DIFF
--- a/libs/langgraph-store-mongodb/CHANGELOG.md
+++ b/libs/langgraph-store-mongodb/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ## Changes in version 0.2.1 (Unreleased)
 
-- Fixes duplicate-key collisions when two items share a namespace component and the same key (INTPYTHON-948). The unique index is now created on a denormalized `namespace_str` field (e.g. `"users/alice/preferences"`) instead of the `namespace` array, which was creating a multikey index in MongoDB.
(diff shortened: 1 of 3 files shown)


caseyclements (commented):
Hi @aclark4life. Have you tested the logic that upgrades an existing collection from an index on [a,b,c] to one on a/b/c?

aclark4life:
> Hi Jeffrey 'Alex' Clark (@aclark4life). Have you tested the logic that upgrades an existing collection from an index on [a,b,c] to one on a/b/c?

No, probably should be an integration test for this.

caseyclements (approved):
LGTM. 
Created INTPYTHON-957 to add one additional test that I'll finish tomorrow.

PR #382, opened 2026-05-06
INTPYTHON-957: test upgrade path from legacy multikey index

caseyclements:
[INTPYTHON-957](https://jira.mongodb.org/browse/INTPYTHON-957)

## Summary

Adds regression coverage for the index-upgrade path in `MongoDBStore.__init__` introduced by [INTPYTHON-948 / #375](https://github.com/langchain-ai/langchain-mongodb/pull/375). That PR replaced a unique compound index on the `namespace` array (which produced multikey collisions) with a unique compound index on a denormalized `namespace_str` field, and added in-place migration logic for collections created under the old schema. The migration logic itself was untested — the existing `store` fixture wipes the collection and drops all indexes before each test, so no test exercised the upgrade branch. This PR closes that gap.

## Changes in this PR

Three new tests in `libs/langgraph-store-mongodb/tests/unit_tests/test_store.py`, plus a dedicated `legacy_collection` fixture (using its own `long_term_memory_legacy` collection so it can't perturb or be perturbed by the shared `store` fixture):

- **`test_upgrade_from_legacy_index`** — sets up a collection with the legacy unique `(namespace, key)` multikey index and pre-existing documents lacking `namespace_str`, then constructs `MongoDBStore` and asserts: every document is backfilled with `namespace_str = "/".join(namespace)`, the legacy index is dropped, a unique `(namespace_str, key)` index now exists, reads return the original values, and a `put` against a legacy `(namespace, key)` updates rather than inserts.
- **`test_upgrade_with_conflicting_non_unique_index`** — creates a non-unique `(namespace_str, key)` index up front, then asserts init drops it and replaces it with the unique variant. Covers the explicit branch at `base.py:218–224`.
- **`test_upgrade_idempotent`** — re-initializes `MongoDBStore` against an already-migrated collection and asserts no document or index changes between the two initializations.

No production code is changed.

## Test Plan

- All three new tests pass against a local Atlas instance: `uv run pytest tests/unit_tests/test_store.py -k upgrade`.
- Full `test_store.py` (12 tests, 9 pre-existing + 3 new) passes: `uv run pytest tests/unit_tests/test_store.py`.
- One side-effect surfaced during test development, flagged here for visibility (not addressed in this PR): `base.py:230` drops the legacy index by key spec, which pymongo resolves to the auto-generated name `namespace_1_key_1`. Collections whose legacy index was created with a custom name would not match, and the drop would raise `OperationFailure: index not found`. Real-world risk is low because PR #375 created the index without a custom name, but worth a separate ticket if we want defense-in-depth.

### Checklist for Author

- [ ] Did you update the changelog (if necessary)? — N/A, tests only
- [x] Is the intention of the code captured in relevant tests? — these tests *are* the intent

### Checklist for Reviewer

### DIFF
--- a/libs/langgraph-store-mongodb/tests/unit_tests/test_store.py
+++ b/libs/langgraph-store-mongodb/tests/unit_tests/test_store.py
@@ -4,6 +4,7 @@ from collections.abc import Generator
 from datetime import datetime
 
 import pytest
+from bson import SON
 from langgraph.store.base import (
     GetOp,
     Item,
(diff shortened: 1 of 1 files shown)


aclark4life (approved):
LGTM! Maybe add typing to match the other tests.

PR #415, opened 2026-07-24
feat: add langchain-mongodb-deepagents-vfs package

anujbpanchal:
## Summary

JIRA issue: [INTPYTHON-1028](https://jira.mongodb.org/browse/INTPYTHON-1028)

Adds a new library, **`langchain-mongodb-deepagents-vfs`**: a MongoDB Atlas-backed
virtual filesystem search adapter for [LangChain DeepAgents](https://github.com/langchain-ai/deepagents).

It implements DeepAgents' `BackendProtocol` and splits work by operation:

- **Search operations** (`grep`, `glob`, `ls`) are routed through MongoDB Atlas —
  vector search, full-text search, and hybrid `$rankFusion`.
- **File operations** (`read`, `write`, `edit`, `upload_files`, `download_files`)
  are forwarded directly to S3.

A change-detection sync layer (initial sync + a polling/SQS watcher) keeps the
Atlas search index in step with the S3 bucket.

## What's included

- `langchain_mongodb_deepagents_vfs/` — backend, S3 adapter, search router,
  chunker, embedder, index manager, sync + watchers (polling / SQS).
- Unit / integration / e2e test suites (unit + integration run without real
  credentials via `mongomock` + `moto`).
- `pyproject.toml`, `justfile`, `README.md`, `CHANGELOG.md` following the
  existing per-lib conventions.
- Registered in `.github/scripts/check_diff.py` `LIB_DIRS` and the root
  `pyproject.toml` `[tool.uv.sources]`.

## Local verification

Run from `libs/langchain-mongodb-deepagents-vfs`:

| Check | Result |
|---|---|
| `ruff check` | ✅ pass |
| `ruff format --check` | ✅ pass |
| `mypy --strict` | ✅ pass (16 source files) |
| Unit tests | ✅ 99 passed |
| `uv lock --check` | ✅ in sync |
| Wheel build + `twine check` | ✅ pass (`test_distribution.py`) |

## ⚠️ Needs a maintainer decision: Python 3.10 CI matrix

This package depends on `deepagents`, which requires `>=3.11,<4.0`. Its
`requires-python` is therefore `>=3.11` and it **cannot** support Python 3.10
like the other three libs.

The shared CI matrix in `.github/workflows/_test.yml` and `_lint.yml` is
hardcoded to `["3.10", "3.13"]` for **every** lib, so the 3.10 job will fail
this package at the install step (before any test runs):

```
$ uv sync --frozen --python 3.10
error: The requested interpreter resolved to Python 3.10.20, which is
incompatible with the project's Python requirement: `>=3.11`
# exit code 2
```

(The 3.13 job installs and passes cleanly.)

I did **not** touch the shared workflow files, since the fix is a repo-wide
policy call. Two options:

1. **Per-library Python matrix** — test this lib on 3.11+ only, leave the others
   on 3.10 + 3.13.
2. **Drop 3.10 repo-wide** — bump the minimum to 3.11 everywhere (breaking for
   the other libs' 3.10 users).

Happy to implement whichever you prefer.

## Notes / smaller items

- E2E MongoDB command-log fixture now writes to pytest's session tmp dir instead
  of the CWD, so the "tests did not create any additional files" CI gate stays
  green; added `mongo-e2e-*.log` to `.gitignore` as a backstop.
- `CHANGELOG.md` uses a simplified heading style vs. the sibling
  `## Changes in version X (date)` convention — happy to align if preferred.
- No `py.typed` yet (matches `langgraph-store-mongodb`); can add if you want type
  info shipped to consumers.


alexbevi:
@caseyclements FYI (this is ported over from https://github.com/mongodb-developer/MongoDB-LangChain-DeepAgents-VFS-Adapter)

caseyclements:
Thanks for your contribution @anujbpanchal! We've created a JIRA ticket [INTPYTHON-1028](https://jira.mongodb.org/browse/INTPYTHON-1028) to track it.

libs/langchain-mongodb-deepagents-vfs/langchain_mongodb_deepagents_vfs/search.py
```diff
@@ -0,0 +1,407 @@
...
+                    "index": "fulltext_search_content_filename",
+                    "wildcard": {
+                        "query": pattern,
+                        "path": "source_path",
+                        # allowAnalyzedField required: source_path uses lucene.keyword analyzer
+                        # (single-token), so wildcard matches against the full path string
+                        "allowAnalyzedField": True,
+                    },
+                }
+            },
```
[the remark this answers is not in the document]
anujbpanchal: The docstring is wrong and I've fixed it, but the stated reasoning isn't quite right — there's no divergence between the two paths. `_glob_atlas` runs the wildcard against `source_path` and `_glob_regex` does `fnmatch.fnmatch(sp, pattern)` where `sp` is also `source_path`, so both implementations agree. Only the docstring was out of date.

There is a real semantics question here though: `fnmatch` lets `*` cross `/`, so `glob("*.pdf")` matches `docs/nested/x.pdf` — unlike shell or pathlib.glob, where `*` stops at a separator. Since agents calling this tool will likely assume shell semantics, I've documented the recursive behaviour explicitly. Happy to switch to shell-style matching instead if that's the preferred style.

libs/langchain-mongodb-deepagents-vfs/tests/fixtures/sample.txt
```diff
@@ -0,0 +1,10 @@
+Line 1: The quick brown fox jumps over the lazy dog.
```
caseyclements: Is this file used? It's mentioned in `confest.FIXTURES_DIR` but that doesn't appear used. 
"sample.txt" is mentioned in text_chunker, but does it actually use the data?
It's an unusual API, the `chunker.chunk(path: str, data: bytes)`. Will this often be `chunk(path, open(path, 'rb').read())`? It's also asking for trouble naming a variable path because it is imported from os so often, and also the chunker should probably take str or pathlib.Path

libs/langchain-mongodb-deepagents-vfs/langchain_mongodb_deepagents_vfs/embedder.py
```diff
@@ -0,0 +1,178 @@
+"""Embedding module — single point of contact with the embedding provider.
+
+Provider is selected at construction time via the ``EMBEDDING_PROVIDER`` env var
+(default: ``openai``).  Supported values:
+
+  openai   — OpenAIEmbeddings (requires OPENAI_API_KEY)
+             Default model: text-embedding-3-small @ 1024 dims
+  bedrock  — BedrockEmbeddings via langchain-aws (uses boto3 credential chain)
+             Default model: amazon.titan-embed-text-v2:0 @ 1024 dims
+
```
caseyclements, line 19: We need to add voyageai here, and autoembedding @alexbevi

libs/langchain-mongodb-deepagents-vfs/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+| `EMBEDDING_PROVIDER` | Default model | Required credential |
+|---|---|---|
+| `bedrock` (default) | `amazon.titan-embed-text-v2:0` | boto3 credential chain (IAM role, `~/.aws/credentials`, etc.) |
+| `openai` | `text-embedding-3-small` | `OPENAI_API_KEY` env var |
+
+Pass a `LangChain Embeddings` instance directly to `embedding_model` to bypass provider lookup entirely.
+
+## Design Decisions
+
```
caseyclements, line 151: Would you add the following?

## Why not `MongoDBStore` + `StoreBackend`?

DeepAgents already ships [`StoreBackend`](https://github.com/langchain-ai/deepagents), which implements the same `BackendProtocol` on top of LangGraph's `BaseStore`. Since `langgraph-store-mongodb`'s `MongoDBStore` *is* a `BaseStore`, this works today and needs no new package:

```python
from deepagents.backends import StoreBackend
from langgraph.store.mongodb import MongoDBStore

backend = StoreBackend(store=MongoDBStore.from_conn_string(...))
```

The two are close cousins — LangGraph's Store is already filesystem-shaped (hierarchical namespaces, keys, prefix search), and `StoreBackend` maps a file path to a single Store key, much as this package maps it to `source_path`. So the difference is not the data model; it is **where search runs** and **how large a file can be**.

Note that `MongoDBStore` itself is not the limiting factor on search: `MongoDBStore.search(query=...)` performs Atlas Vector Search, with optional reranking. The gap is that `StoreBackend` never calls it that way — its `grep` loads items and matches literally in Python, so the Store's semantic search goes unused. The rows below therefore describe the pairing as `StoreBackend` wires it up, not the ceiling of what `MongoDBStore` can do.

| | `MongoDBStore` + `StoreBackend`                                                                          | This package |
|---|----------------------------------------------------------------------------------------------------------|---|
| `grep` / `glob` execution | `StoreBackend` fetches every item in the namespace, then filters in Python                               | runs as an aggregation inside MongoDB; only matches come back |
| `grep` matching | literal substring — `StoreBackend` does not pass `query` to `MongoDBStore.search()`, so no vector search | hybrid `$rankFusion` — full-text **and** vector, so natural-language queries work |
| Unit of storage | one Store item per file, read and written whole                                                          | file chunked into 512-token documents |
| Large files | whole file must fit in memory on every `grep`                                                            | only matching chunks are returned; `line_start` gives real line numbers |
| Where bytes live | in MongoDB, as the Store value                                                                           | in S3; MongoDB holds only chunks and embeddings |
| Source of truth | MongoDB                                                                                                  | the S3 bucket — objects written by other tools are picked up automatically |

The chunking is the part that cannot be retrofitted onto `StoreBackend`: it depends on one Store item per file path, and splitting a file into many items would break that identity. Chunking is also what makes semantic `grep` useful, since embedding a whole document into one vector loses the passage-level detail an agent is searching for.

**Use `MongoDBStore` + `StoreBackend`** for an agent scratchpad — modest numbers of small, agent-authored files, where exact-match `grep` is fine and you want writes and reads in one system.

**Use this package** when the corpus is large, pre-existing, and lives in S3 (documents, PDFs, spreadsheets), and the agent needs to find things by meaning rather than by exact string.

### A note on search freshness

Neither option gives read-your-writes for search. Atlas Search and Vector Search are eventually consistent by design: `mongod` replicates to the `mongot` process, which indexes asynchronously, so any newly written document becomes searchable a short time after the write acknowledges. Point reads are unaffected — `read` here goes straight to S3, and a `MongoDBStore.get()` (or a `search()` with only a `filter` and no `query`) is a normal MongoDB query.

This package adds two further hops on top of that: a `write` lands in S3, and the object is only chunked and embedded once the watcher notices it — up to the poll interval (10s default) with `PollingWatcher`, or near real-time with `SQSWatcher`. Budget for `write` → `grep` lag on the order of the watcher interval plus Atlas indexing time, and don't rely on a file being greppable immediately after writing it.

caseyclements:
Hi @anujbpanchal. 

1. Please resolve the issues mentioned about `glob`.
2. Update the README to differentiate this package from MongoDBStore. Draft provided.
3. Update the minimum Python version from 3.10 to 3.11 in the global test matrix. I'll help out if anything should arise by creating a separate PR.

caseyclements (changes requested):
Suggestions above.

libs/langchain-mongodb-deepagents-vfs/langchain_mongodb_deepagents_vfs/search.py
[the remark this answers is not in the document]
caseyclements: > The docstring is wrong and I've fixed it, but the stated reasoning isn't quite right — there's no divergence between the two paths. `_glob_atlas` runs the wildcard against `source_path` and `_glob_regex` does `fnmatch.fnmatch(sp, pattern)` where `sp` is also `source_path`, so both implementations agree. Only the docstring was out of date.
> 
> There is a real semantics question here though: `fnmatch` lets `*` cross `/`, so `glob("*.pdf")` matches `docs/nested/x.pdf` — unlike shell or pathlib.glob, where `*` stops at a separator. Since agents calling this tool will likely assume shell semantics, I've documented the recursive behaviour explicitly. Happy to switch to shell-style matching instead if that's the preferred style.

I would use shell semantics if it isn't a big lift. You've stated the reason - we want agents to get things right the first time.

caseyclements:
Hi @anujbpanchal Almost there. Small CI failures. 
e.g. `botocore.exceptions.NoRegionError: You must specify a region.`, some python3.10 stuff

caseyclements (approved):
LGTM!

PR #418, opened 2026-07-30
INTPYTHON-981 Fix AutoEmbeddings index configuration

gulbaki:
[INTPYTHON-981]

## Summary

Fix vector index configuration when using `AutoEmbeddings` without explicitly providing dimensions or similarity.

## Changes in this PR

- Infer the required AutoEmbeddings index settings.
- Preserve existing defaults for manual embeddings.
- Add regression tests and update the AutoEmbeddings integration test.
- Update the changelog and docstring.

## Test Plan

- Added five unit tests covering AutoEmbeddings and manual embeddings.
- Unit tests, mypy, Ruff, and pre-commit checks pass.
- The integration test was not run locally because MongoDB Community with Search was unavailable.

### DIFF
--- a/libs/langgraph-store-mongodb/langgraph/store/mongodb/base.py
+++ b/libs/langgraph-store-mongodb/langgraph/store/mongodb/base.py
@@ -119,7 +119,8 @@ class VectorIndexConfig(IndexConfig, total=False):
     It is designed to have one vector per document.
 
     NOTE: If using AutoEmbeddings, the vectors are not explicitly stored in the Collection.
-    Set dims to -1, relevance_score_fn to None.
+    The factory function automatically configures dimensions and similarity for
+    auto-embedding indexes.
     The embedding_key will not store vectors. Instead, it will be the texts to be embedded.
     """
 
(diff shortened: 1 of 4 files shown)


gulbaki:
Hi @aclark4life @caseyclements this PR fixes the AutoEmbeddings vector index configuration in langgraph-store-mongodb and adds regression tests. Could you please take a look when you get a chance? Thanks!

caseyclements:
Thank you for contributing @gulbaki! We'll take a look. @sleepyStick

libs/langgraph-store-mongodb/langgraph/store/mongodb/base.py
```diff
@@ -139,28 +140,46 @@ class VectorIndexConfig(IndexConfig, total=False):
...
+            Defaults to cosine for manual embeddings and must be omitted for
+            AutoEmbeddings.
         embedding_key: Name of the field used in the collection to store vectors.
         filters: List of (possibly nested) fields to index allowing filtering.
 
     Returns: VectorIndexConfig to be passed to MongoDBStore constructor.
     """
 
+    if embed is None:
+        raise ValueError("embed is required.")
```
sleepyStick, line 169: if we're raising a value error when embed is None, doesn't that mean that the param isn't `Optional`?

sleepyStick (changes requested):
For the most part i think it looks good? Just one question about the type of a param.

libs/langgraph-store-mongodb/langgraph/store/mongodb/base.py
sleepyStick: fixed in [1b83547](https://github.com/langchain-ai/langchain-mongodb/pull/418/commits/1b835477d603b2a3ad0a50011c0a0d5e87414143)

sleepyStick:
Because `embed` isn't actually optional, `dims` cant take a default value of `None` (assuming we want to keep the current param order of `dims` and then `embed`. If we're okay with breaking that, then `dims` can have a default value of `None` again.

libs/langgraph-store-mongodb/langgraph/store/mongodb/base.py
```diff
@@ -143,24 +144,37 @@ def create_vector_index_config(
     embed: Union[Embeddings, EmbeddingsFunc, AEmbeddingsFunc, str],
     fields: Optional[list[str]] = None,
     name: str = "vector_index",
-    relevance_score_fn: Literal["euclidean", "cosine", "dotProduct", None] = "cosine",
+    relevance_score_fn: Literal["euclidean", "cosine", "dotProduct", None] = None,
```
caseyclements: For backwards-compatibility reasons, I would like to keep cosine as the default. 
The docstring below looks good.

libs/langgraph-store-mongodb/tests/integration_tests/test_autoembedded_index.py
```diff
@@ -71,8 +69,7 @@ def test_filters(collection: Collection) -> None:
 
     index_config = create_vector_index_config(
         name=INDEX_NAME,
-        dims=DIMENSIONS,
-        relevance_score_fn=None,
```
caseyclements: `relevance_score_fn=None,` should remain.

PR #425, opened 2026-08-18
fix(deepagents-vfs): PollingWatcher survives callback exceptions

caseyclements:
## Summary
- `_loop()` previously only guarded the `list_keys()` call inside `_poll()`; an exception raised from `on_created`/`on_updated` (e.g. a raw pymongo error, not an `AdapterError`) propagated out of `_poll` and killed the background polling thread with no signal to the caller.
- `_loop()` now wraps each poll iteration in a broad exception handler, logs the failure, and records it on `watcher.last_error` so callers can detect a watcher that has silently stopped syncing.

## Test plan
- [x] Added `test_loop_survives_callback_exception_and_records_health`, which crashes `on_created` and asserts the loop survives and `last_error` is set (watched it fail before the fix).
- [x] `uv run pytest tests/unit_tests/test_watcher.py -v` — all pass.


blink1073 (approved):
LGTM

PR #426, opened 2026-08-18
fix(deepagents-vfs): surface fatal init failures from grep/glob/ls

caseyclements:
## Summary
- `_wait_ready()` only blocked on the ready event. A failed index-provisioning step or a failed initial sync was already recorded in `init_errors`, but `grep`/`glob`/`ls` proceeded anyway and returned empty results indistinguishable from "nothing matched" — the package's worst failure mode: confidently wrong empty results.
- Fatal init failures (index provisioning or initial sync raising outright) are now tracked separately from a watcher-start failure, which doesn't invalidate already-synced data. `_wait_ready()` raises `AdapterError` (`E1004_INDEX_PROVISION_FAILED` / `E1005_INITIAL_SYNC_FAILED`) in the fatal case, which the `adapter_boundary` decorator turns into a proper error DTO for callers.

## Test plan
- [x] Added `TestSearchSurfacesInitFailures`: `grep` errors when index provisioning failed, `glob` errors when initial sync failed, `ls` still succeeds when only the watcher failed to start (watched these fail before the fix).
- [x] `uv run pytest tests/unit_tests/test_backend.py -v` — all pass.


blink1073 (approved):
LGTM

PR #430, opened 2026-08-19
INTPYTHON-1060 fix(deepagents-vfs): bound the fulltext branch of hybrid grep's $rankFusion

caseyclements:
## Summary
- `SearchRouter._grep_hybrid`'s `$rankFusion` combines a vector branch and a fulltext branch, but only the vector branch was actually bounded: `vector_search_stage()` bakes `limit`/`numCandidates` directly into the `$vectorSearch` stage itself, while the hand-rolled fulltext branch (`$search` + optional `$match`) had no `$limit` of its own — only a single `$limit` trailing the entire `$rankFusion` combination.
- A common search term against a large collection let the fulltext branch rank an unbounded candidate set before that trailing limit ever got a chance to trim it, and the two branches feeding the same RRF combination were sized asymmetrically.
- Adds an explicit `{"$limit": top_k}` to the fulltext branch, matching the vector branch's `top_k` (`grep_limit * 2`), so both branches are bounded the same way before combination.

Kept the existing hand-rolled `$search` stage rather than switching to the `text_search_stage()` helper already used elsewhere in this file — that helper defaults to emitting an extra `$set: {score: ...}` stage, which is wasted work here since `$rankFusion` combines by rank order, not raw `$search` score, and nothing downstream reads a `score` field.

Note: this does not address the separate, larger question of whether `path`/`glob` scoping should be pushed into the `$search`/`$vectorSearch` stages themselves rather than applied as a post-search `$match` — that's tracked separately (see `jira_grep_hybrid.txt` for the fuller writeup) as a decision that needs verifying Atlas Search's `compound.filter` support, not something to bundle into this minimal fix.

## Test plan
- [x] Added `TestGrepHybridPipelineShape` in `test_search_router.py`: captures the constructed `$rankFusion` pipeline via a mocked collection and asserts the fulltext branch has its own `$limit`, matching the vector branch's `top_k`.
- [x] `uv run pytest tests/unit_tests tests/integration_tests` — all pass (156 tests).

### DIFF
--- a/libs/langchain-mongodb-deepagents-vfs/langchain_mongodb_deepagents_vfs/search.py
+++ b/libs/langchain-mongodb-deepagents-vfs/langchain_mongodb_deepagents_vfs/search.py
@@ -271,11 +271,18 @@ class SearchRouter:
         pre_filter: dict[str, Any] = {"$and": path_filter} if path_filter else {}
 
         if query_vector is not None:
+            # Both branches feed the same combination step, so both need a
+            # comparable bound: $vectorSearch already caps itself via its own
+            # limit/numCandidates fields, but $search has no such built-in cap
+            # — without an explicit $limit here, the fulltext branch could
+            # rank an unbounded number of candidates before the single $limit
+            # trailing the whole $rankFusion ever gets a chance to trim it.
+            top_k = self._grep_limit * 2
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
LGTM

PR #431, opened 2026-08-20
INTPYTHON-1054 Fix CI failures for py3.14 coming from pydantic

caseyclements:
[INTPYTHON-1054](https://jira.mongodb.org/browse/INTPYTHON-1054)

## Summary

Prep both `langchain-mongodb` and `langchain-mongodb-deepagents-vfs` for Python 3.14 by raising their `pydantic` floor to `>=2.13.0`, plus three further floor bumps in `langchain-mongodb-deepagents-vfs` (`tiktoken`, `pytest-asyncio`, `mongomock`) — see "Additional deepagents-vfs floors" below.

`pydantic-core` versions pulled in by `pydantic < 2.13` are Rust extensions built via PyO3, and PyO3 0.22.2 (what those older `pydantic-core` builds used) has no support for the Python 3.14 ABI at all — there's no prebuilt wheel, and building from source fails outright:

```
error: the configured Python interpreter version (3.14) is newer than
PyO3's maximum supported version (3.13)
💥 maturin failed
```

Both packages currently declare `langchain-core>=0.2.0`/`langchain>=1.0`, which only constrain `pydantic<3.0.0,>=2.7.4` transitively — too low to guarantee a Python 3.14-compatible `pydantic-core` gets resolved. Pinning `pydantic>=2.13.0` explicitly (verified to resolve to a build with an actual 3.14 wheel, no compilation needed) closes that gap.

Note: this was initially suspected to be `langchain-mongodb`'s `filterwarnings = ["error", ...]` pytest setting turning a `langchain-core` pydantic-v1-import deprecation warning into a test failure. That mechanism is real for `langchain-mongodb`, but reproducing the actual failure showed it's unrelated — it's a hard build-time incompatibility in `pydantic-core`, not a warning. `langchain-mongodb-deepagents-vfs` doesn't even have a `filterwarnings` override, so that explanation couldn't have applied there regardless.

## Additional deepagents-vfs floors

Three further floors in `libs/langchain-mongodb-deepagents-vfs/pyproject.toml` were too low. None were catchable by CI, because the "minimum dependency versions" step is a no-op — `uv run` re-resolves to highest before pytest starts (fixed separately in #433 / INTPYTHON-1066).

- **`tiktoken>=0.8.0` → `>=0.12.0`.** 0.8.0 ships no cp314 wheel, so a 3.14 install falls back to a source build of its Rust extension and fails with `can't find Rust compiler`. 0.12.0 is the first release with cp314 wheels. Same class of failure as the pydantic/PyO3 issue above; the existing inline comment (`# first release with cp313 wheels`) was exactly the right reasoning, one Python version out of date.
- **`pytest-asyncio>=0.23.0` → `>=0.23.8`.** With `pytest==8.0.0` (our floor), 0.23.0 and 0.23.3 crash during collection with `INTERNALERROR: 'Package' object has no attribute 'obj'` and run **zero tests**. 0.23.8 is the lowest that works.
- **`mongomock>=4.1.0` → `>=4.3.0`.** 4.1.1 and 4.1.2 import `pkg_resources`, absent on modern Python, so `conftest`'s `importorskip` silently skipped **29 of 131 tests**. Only those three releases exist at or above the floor; 4.3.0 is the lowest that works.

Verified for deepagents-vfs: 3.14 `lowest-direct` sync succeeds; the 3.11 `lowest-direct` suite goes from `INTERNALERROR` / 0 tests to **131 passed, 0 skipped**; the highest-version suite is still 131 passed; mypy clean.

This package is not yet published to PyPI, so all three bumps are free of downstream-compatibility cost.

## Changes in this PR

- `libs/langchain-mongodb/pyproject.toml`: add `pydantic>=2.13.0` to `dependencies`.
- `libs/langchain-mongodb-deepagents-vfs/pyproject.toml`: add `pydantic>=2.13.0` to `dependencies`.
- Regenerated `uv.lock` for both packages and the top-level lockfile to reflect the new floor (`pydantic` resolves to `2.13.4`).

No source code changes — this is a dependency floor bump plus lockfile regeneration only.

## Test Plan

- Reproduced the actual failure directly: in a clean Python 3.14 venv, `pydantic==2.9.2` (pulling `pydantic-core==2.23.4`) fails to build via PyO3/maturin with the ABI-version error above; `pydantic>=2.13.0` installs instantly from a prebuilt wheel with no compilation step.
- Confirmed no test-suite changes were needed — this is a resolution/build-compatibility fix, not a behavior change, and no new code paths are introduced.
- Existing unit/integration suites for both packages pass unchanged with the new `pydantic` floor.

### Checklist for Author

- [x] If there are new TODOs, has a related JIRA ticket been created?
- [x] Has a MongoDB Employee run [the patch build of this PR](https://github.com/mongodb-labs/ai-ml-pipeline-testing?tab=readme-ov-file#running-a-patch-build-of-a-given-pr)?

### Checklist for Reviewer


caseyclements:
[Passing patch](https://spruce.corp.mongodb.com/version/6a88c9b5f6b0090007431f35/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

blink1073 (approved):
LGTM

PR #435, opened 2026-08-23
Prep langchain-mongodb 0.12.0 release

caseyclements:
Release prep for `langchain-mongodb` 0.12.0, per [RELEASE.md](https://github.com/langchain-ai/langchain-mongodb/blob/main/RELEASE.md). **Not the release itself** — this only bumps the version and dates the changelog so the release workflow can be dispatched afterwards.

## Summary

- `libs/langchain-mongodb/pyproject.toml`: `version = "0.11.0"` → `"0.12.0"`
- `libs/langchain-mongodb/CHANGELOG.md`: dates the existing 0.12.0 section `(XXXX/XX/XX)` → `(2026/08/24)`

Minor bump is appropriate — 0.12.0 carries new features, not just fixes:

- `MongoDBStructuredQueryTranslator` ISO 8601 date/datetime conversion fix
- Auto-embeddings support in `MongoDBAtlasHybridSearchRetriever`, `MongoDBAtlasParentDocumentRetriever`, `MongoDBAtlasSelfQueryRetriever`
- Native `$rerank` support across `MongoDBAtlasVectorSearch` and all retrievers

## Changes in this PR

Two single-line changes, no source changes. Full diff is 2 files, 2 insertions, 2 deletions.

## Test Plan

`main` was verified this morning at `006027e` (the commit this branches from) across the full CI matrix, simulating the minimum-version step from #433:

| Lib | py3.11 | py3.13 |
|---|---|---|
| `langchain-mongodb` | 75 passed | 75 passed |
| `langchain-mongodb-deepagents-vfs` | 134 passed | 134 passed |
| `langgraph-checkpoint-mongodb` | 17 passed | 17 passed |
| `langgraph-store-mongodb` | 23 passed | 23 passed |

No behavioural change in this PR itself; CI on the branch is the gate.

## Two questions for the reviewer

**1. The changelog does not mention the dependency floor changes from #431 and #432.** These are user-visible and can break installs for anyone pinned below the new floors:

| Dependency | Change |
|---|---|
| `langchain-core` | `>=1.2.5` → `>=1.4.2` (runtime) |
| `pydantic` | newly declared, `>=2.13.0` |
| `numpy` | `>=1.26`, plus `>=2.3.2` on Python 3.14 via environment marker |
| `typing-extensions` | `>=4.12.2` → `>=4.13.0` (dev only) |

Worth a changelog line before release, since a raised runtime floor is the kind of change a consumer needs to see. Suggested wording:

> - Raise minimum dependency versions for Python 3.14 support: `langchain-core>=1.4.2`, `pydantic>=2.13.0`, and `numpy>=2.3.2` on Python 3.14 (unchanged at `>=1.26` on 3.11–3.13).

**2. The changelog date is `2026/08/24` (tomorrow).** `RELEASE.md` says "including today's date". Intentional if the release is planned for tomorrow — just confirming it isn't a slip.

### Checklist for Author

### Checklist for Reviewer


blink1073 (approved):
LGTM

blink1073:
Ah, I think we should wait on #434 and then update this PR and its lock.

caseyclements:
## Proposal: drop TestPyPI from the release process

Not part of this PR — posting here for visibility while the release is being reviewed. TestPyPI is being deprecated upstream and MongoDB has begun removing it elsewhere; it is also currently a hard blocker for the first `langchain-mongodb-deepagents-vfs` release, since a pending publisher must be registered on TestPyPI under an account that may not match the PyPI one.

Below is what would have to change. **Nothing is committed.**

### Scope

| File | Change |
|---|---|
| `.github/workflows/_test_release.yml` | **delete entirely** (89 lines) |
| `.github/workflows/_release.yml` | remove one job, three `needs:` entries, rework two steps |
| `TRUSTED_PUBLISHING.md` | drop step 2; only one publisher would be needed |
| `RELEASE.md` | no change — it never mentions TestPyPI |

### 1. Delete the `test-pypi-publish` job

```yaml
  test-pypi-publish:            # _release.yml:79-88 — delete
    needs:
      - build
    uses:
      ./.github/workflows/_test_release.yml
    ...
```

### 2. Drop it from three `needs:` lists

`pre-release-checks` (:93), `publish` (:196), `mark-release` (:239) each list `- test-pypi-publish`. Remove that line from all three; `- build` and `- pre-release-checks` still give the correct ordering.

### 3. Rework `pre-release-checks` — the substantive part

This job exists to prove the **built artifact** installs and passes tests the way a consumer would receive it. TestPyPI was only the delivery mechanism. The replacement is to install the wheel straight from the `dist` artifact the `build` job already uploads.

Note the job does **not** currently download that artifact, so a `download-artifact` step has to be added.

```yaml
      - uses: actions/download-artifact@v8          # NEW
        with:
          name: dist
          path: ${{ inputs.working-directory }}/dist/

      - name: Import built package                  # replaces "Import published package"
        shell: bash
        env:
          PKG_NAME: ${{ needs.build.outputs.pkg-name }}
        run: |
          # Install the built wheel; dependencies resolve from real PyPI.
          uv run pip install dist/*.whl

          IMPORT_NAME="${PKG_NAME//-/_}"
          [[ "$PKG_NAME" == langgraph* ]] && IMPORT_NAME="${PKG_NAME//-/.}"
          uv run python -c "import $IMPORT_NAME; print(dir($IMPORT_NAME))"

      - uses: extractions/setup-just@...
      - name: Import test dependencies
        run: just install

      # `just install` puts an editable install of the local source in place;
      # overwrite it so the tests run against the built artifact.
      - name: Reinstall built package over the editable install
        run: uv run pip install --force-reinstall --no-deps dist/*.whl
```

The two retry loops (`|| ( sleep 5 && ... )`) can go — they only existed because TestPyPI has propagation delay. A local file does not.

### Trade-offs

**Lost:** a genuine round-trip through a warehouse instance — upload, index, download. That implicitly validated that PyPI would accept the artifact's metadata *before* the real upload. Metadata rejection would now surface in the `publish` job instead, i.e. later, after `pre-release-checks` has already passed.

Concretely relevant here: the `Metadata-Version: 2.5` question from #427. PyPI does accept 2.5 (verified — `hatchling` 1.32.0 and `langsmith` 0.11.1 are both live with it), so this is not a live risk, but it is the class of problem the TestPyPI round-trip was catching.

**Gained:**
- No second registry account to keep in sync — the immediate blocker for deepagents-vfs.
- Dependencies resolve from **real** PyPI rather than a TestPyPI/PyPI hybrid, so `pre-release-checks` is a more faithful simulation of a consumer install than it is today.
- Removes `skip-existing: true`, commented in `_test_release.yml` as *"only for CI use and extremely dangerous otherwise"*.
- One less OIDC publisher to configure per new package, and a shorter pipeline.

**Optionally recovering the lost check:** add `uvx twine check dist/*` to the `build` job. That validates metadata locally without needing an index — cheap, and covers most of what the round-trip was buying.

### Suggested sequencing

Its own PR, after this release lands. It touches `.github`, so `check_diff.py` will run all four libs — and it should not be tangled up with a release-prep diff that is currently two lines.

caseyclements:
[Passing build](https://spruce.corp.mongodb.com/version/6a8cc955dbafff00073b6baf/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

PR #438, opened 2026-08-24
INTPYTHON-1074 Drop the langchain-mongodb dependency from deepagents-vfs

caseyclements:
[INTPYTHON-1074](https://jira.mongodb.org/browse/INTPYTHON-1074)

## Summary

`langchain-mongodb` was used for exactly one thing in `langchain-mongodb-deepagents-vfs`: a single import of `text_search_stage` at `search.py:16`, called once in the no-vector fallback branch of `_grep_hybrid`.

The same file already imported the sibling helper from the other package:

```python
from langchain_mongodb.pipelines import text_search_stage   # ← removed
from pymongo_search_utils import vector_search_stage
```

So two closely related functions were pulled from two different packages, and one of those packages existed in the dependency tree solely to supply it.

## Why the swap is safe

The two `text_search_stage` implementations are separate code, not a re-export — but diffing the source shows the only differences are typing syntax:

```diff
-    search_field: Union[str, List[str]],
+    search_field: str | list[str],
-    limit: Optional[int] = None,
+    limit: int | None = None,
-) -> List[Dict[str, Any]]:
+) -> list[dict[str, Any]]:
```

Same parameters, same defaults, same emitted stages (`$search` → optional `$match` → optional `$set` score → optional `$limit`). For contrast, `vector_search_stage` genuinely *is* the same object in both packages (`langchain_mongodb` re-exports it from `pymongo_search_utils`), which is why only `text_search_stage` needed attention.

**No floor change needed.** `pymongo-search-utils` is already declared at `>=0.3.0`, and `text_search_stage` has been present since 0.1.0 — verified by installing 0.1.0, 0.2.0 and 0.3.0 and checking for the attribute.

## Impact

Nine packages leave a runtime install:

```
async-timeout, greenlet, importlib-metadata, langchain-classic,
langchain-mongodb, langchain-text-splitters, lark, numpy, sqlalchemy
```

`langchain-mongodb`'s full transitive closure is 50 packages, but most arrive anyway through `deepagents` (which pulls `langchain`, `langchain-core`, `langsmith` and their trees) — so **nine is the honest net figure**, not fifty. `numpy` and `sqlalchemy` are the notable weight.

One nuance: `numpy` still appears in `uv.lock`, required by `langchain-aws`. That is only reachable through the `[bedrock]` extra — a base install no longer pulls it, confirmed below.

## Secondary benefit: removes release-order coupling

While planning the first deepagents-vfs release we had to reason about whether it should ship before or after `langchain-mongodb` 0.12.0, because the `>=0.11.0` floor named a version predating the Python 3.14 dependency fixes (INTPYTHON-1054, INTPYTHON-1065). That question disappears permanently, along with one more sibling floor to keep current.

## Test Plan

- Unit: **134 passed**. Integration: **26 passed**.
- Minimum versions (`--resolution=lowest-direct`, Python 3.11, run with `--no-sync`): **134 passed** — worth checking explicitly, since the resolved graph changes materially.
- `mypy` clean (16 source files); `ruff` and `ruff format` clean.
- Runtime-only install (`uv sync --no-default-groups`) confirms `langchain-mongodb`, `numpy`, `sqlalchemy` and `lark` are all **absent**.
- No `langchain_mongodb` references remain anywhere in the package source.

## Timing

Worth landing **before** the first release. The package is unpublished, so removing a dependency now costs nothing; afterwards it is a visible change to consumers' resolved environments.

### Checklist for Author

### Checklist for Reviewer

### DIFF
--- a/libs/langchain-mongodb-deepagents-vfs/langchain_mongodb_deepagents_vfs/search.py
+++ b/libs/langchain-mongodb-deepagents-vfs/langchain_mongodb_deepagents_vfs/search.py
@@ -13,9 +13,8 @@ import re
 from collections.abc import Iterable
 from typing import Any
 
-from langchain_mongodb.pipelines import text_search_stage
 from pymongo.collection import Collection
-from pymongo_search_utils import vector_search_stage
+from pymongo_search_utils import text_search_stage, vector_search_stage
 from wcmatch.glob import BRACE, GLOBSTAR, globmatch
 
 from langchain_mongodb_deepagents_vfs.dtypes import (
--- a/libs/langchain-mongodb-deepagents-vfs/pyproject.toml
+++ b/libs/langchain-mongodb-deepagents-vfs/pyproject.toml
@@ -19,7 +19,6 @@ dependencies = [
     "boto3>=1.34.0",
     "pymongo>=4.6.1",
     "langchain-core>=0.2.0",
-    "langchain-mongodb>=0.11.0",
     "pydantic>=2.13.0",
     "pymongo-search-utils>=0.3.0",
     "pypdf>=6.15.0",
--- a/libs/langchain-mongodb-deepagents-vfs/uv.lock
+++ b/libs/langchain-mongodb-deepagents-vfs/uv.lock
@@ -546,69 +546,6 @@ wheels = [
 ]
 
 [[package]]
-name = "greenlet"
-version = "3.5.3"
-source = { registry = "https://pypi.org/simple" }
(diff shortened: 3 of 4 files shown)


sleepyStick (approved):

caseyclements:
[Passing build](https://spruce.corp.mongodb.com/version/6a8cc4f8869a4c0007165dea/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

PR #440, opened 2026-08-26
docs: document text_key parameter in langchain-mongodb README

### LINKED ISSUE #211: MongoDBAtlasVectorSearch ambiguous behavior
### Description

I am building a system with MongoDBAtlasVectorSearch and I came across a problem which took me around 3 hours to figure out.

Specifically, when initializing MongoDBAtlasVectorSearch, the class expects your documents to contain a "text" field. After getting docs from mongodb, it filters the output based on whether the doc contains this "text" field. In my case, the field I used for storing text was called "text_clean", so I got empty retrieval outputs even though docs in my collection had embedding field and the query itself was successful. Providing a text_key optional argument to the class solves the problem.

I consider this behavior non-obvious and suggest to either explicitly include text_key argument in the examples in the docs, like 
```
vector_store = MongoDBAtlasVectorSearch(
    collection=MONGODB_COLLECTION,
    embedding=embeddings,
    index_name=ATLAS_VECTOR_SEARCH_INDEX_NAME,
    relevance_score_fn="cosine",
    text_key="the field your collection docs should contain to be included in the final retrieval output !!!!!"
)
```
or disable filtering of retrieval results by text_key by default.

SpiliosDimakopoulos:
Closes #211.

The MongoDBAtlasVectorSearch quickstart example in the README didn't
show the `text_key` parameter. Users whose collections store page
content under a field other than the default "text" (e.g. "text_clean")
got silent, empty `similarity_search` results with no indication of
why, as described in #211.

This adds `text_key="text"` to the example call along with a short
comment explaining what it does and why it matters.

### DIFF
--- a/libs/langchain-mongodb/README.md
+++ b/libs/langchain-mongodb/README.md
@@ -34,6 +34,11 @@ vectorstore = MongoDBAtlasVectorSearch.from_connection_string(
     namespace=DB_NAME + "." + COLLECTION_NAME,
     embedding=OpenAIEmbeddings(model=MODEL_NAME),
     index_name=VECTOR_SEARCH_INDEX_NAME,
+    # The field in your documents that holds the page content. Defaults to
+    # "text" — if your collection stores it under a different field name
+    # (e.g. "text_clean" or "content"), set this explicitly or
+    # similarity_search will silently return no results.
+    text_key="text",
 )
 
 retrieved_docs = vectorstore.similarity_search(


NoahStapp:
#211 was addressed in https://github.com/langchain-ai/langchain-mongodb/pull/220. That PR also makes it immediately visible for users when they may have a mismatching configuration. Closing this as unnecessary, but thank you for the effort!

PR #441, opened 2026-08-26
fix: make date-only query comparisons UTC-aware in MongoDBStructuredQueryTranslator

SpiliosDimakopoulos:
## Problem

`MongoDBStructuredQueryTranslator._convert_dict_to_datetime()` was
producing **naive** datetime objects for ISO 8601 date-only strings
(e.g. `{"date": "2025-01-01", "type": "date"}`).

MongoDB BSON / Atlas Vector Search requires timezone-aware datetimes
for filter comparisons to work correctly. Naive datetimes compare
against stored UTC timestamps incorrectly, causing date-range filters
(e.g. "Movies released before 1960") to silently return wrong results.

## Fix

Attach `tzinfo=timezone.utc` when parsing date-only strings so the
resulting datetime is midnight UTC — consistent with how the `datetime`
branch already handles the `Z` suffix.

Also updates the three existing unit tests that were asserting the old
(incorrect) naive-datetime behaviour, and adds an explicit
`tzinfo is not None` guard to make the contract explicit.

## Changes

- `libs/langchain-mongodb/langchain_mongodb/retrievers/self_querying.py` — one-line fix
- `libs/langchain-mongodb/tests/unit_tests/test_retrievers.py` — corrected assertions

### DIFF
--- a/libs/langchain-mongodb/langchain_mongodb/retrievers/self_querying.py
+++ b/libs/langchain-mongodb/langchain_mongodb/retrievers/self_querying.py
@@ -1,4 +1,5 @@
 import datetime
+from datetime import timezone
 from typing import Any, Dict, Sequence, Tuple, Union
 
 from langchain_classic.chains.query_constructor.schema import (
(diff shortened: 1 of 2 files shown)


NoahStapp (changes requested):
Hi,

Thanks for your contribution! PyMongo already encodes naive datetimes as UTC, described in the official documentation: https://www.mongodb.com/docs/languages/python/pymongo-driver/current/data-formats/dates-and-times/#aware-datetimes. Do you have a specific example where this fails?

PR #458, opened 2026-09-10
INTPYTHON-1075: Remove COMMUNITY_WITH_SEARCH guards now that autoembedding is available on all Cloud, Local, and Community

caseyclements:
[INTPYTHON-1075](https://jira.mongodb.org/browse/INTPYTHON-1075)

## Summary

After [ai-ml-pipeline-testing/pull/120](https://github.com/mongodb-labs/ai-ml-pipeline-testing/pull/120) is merged, Auto-embedded Vector Search Indexes will be available on all three platforms we test against — cloud Atlas, local Atlas, and Community with Search. The `COMMUNITY_WITH_SEARCH` environment variable existed to gate Auto-embedding tests to the one platform that supported them, and is no longer needed.

__This PR should be merged shortly after the other.__ I will include a passing patch build on ai-ml-pipeline-testing pointing to feature branch here. Then I will do the same for this one.

## Changes in this PR

Removed the `COMMUNITY_WITH_SEARCH` guard and the now-unused `import os` from:

- `libs/langchain-mongodb/tests/integration_tests/test_autoembedded_vectorstore.py` — module-level `pytestmark` skip
- `libs/langchain-mongodb/tests/integration_tests/test_retrievers.py` — `skipif` on `test_hybrid_retriever_autoembed`
- `libs/langchain-mongodb/tests/integration_tests/test_parent_document.py` — skip on the `auto` branch of the `embedding_param` fixture
- `libs/langchain-mongodb/tests/integration_tests/test_index.py` — inverted module-level skip
- `libs/langchain-mongodb/tests/unit_tests/test_index.py` — inverted module-level skip
- `libs/langgraph-store-mongodb/tests/integration_tests/test_autoembedded_index.py` — module-level skip

The env var is now gone from the entire repository.

Also in `libs/langchain-mongodb/tests/utils.py`: `PatchedMongoDBAtlasVectorSearch` had two separate wait-for-indexing loops, polling `list_search_indexes()[...]["numDocs"]` for the autoembedding case and `similarity_search` otherwise. Since `similarity_search` handles the autoembedding path, these are collapsed into a single loop and `_is_autoembedding` is no longer branched on here.

Minor: fixed a `semnatic` -> `semantic` typo in the `test_autoembedded_index.py` module docstring.

## Test Plan

This PR is entirely test-side; the change under test is which platforms the existing autoembedding tests run on.

[...]

PR #459, opened 2026-09-11
INTPYTHON-1075: Skip auto-embedding tests where the deployment has no embedding model registered

caseyclements:
[INTPYTHON-1075](https://jira.mongodb.org/browse/INTPYTHON-1075)

## Summary

#458 removed the `COMMUNITY_WITH_SEARCH` guards so the auto-embedding tests run on every platform. On fork PRs they then fail, because `createSearchIndexes` is refused outright:

```
pymongo.errors.OperationFailure: CanonicalModel: voyage-4 not registered yet, supported models are: []
```

The local Atlas container CI starts has no Voyage model registered. Per [ai-ml-pipeline-testing#120](https://github.com/mongodb-labs/ai-ml-pipeline-testing/pull/120), auto-embedding on `mongodb-atlas-local` requires `VOYAGE_API_KEY` forwarded into the container *and* the `:preview` image tag. `scripts/start_local_atlas.sh` passes neither and pins `:latest`, so no GitHub Actions run has auto-embedding on any trigger — and a fork PR gets an empty `secrets` context, so it could never supply the key even if the script forwarded one.

So this is a static property of the deployment, and the script that provisions it now says so.

## Changes in this PR

- `scripts/start_local_atlas.sh` writes `AUTOEMBED_UNSUPPORTED=1` alongside `MONGODB_URI`. No workflow change is needed: `.local_atlas_uri` is already `dotenv-load`ed by both justfiles and cat'd into `$GITHUB_ENV` at `_test.yml:58`. The explanatory comment lives in the script, not the generated file, because `$GITHUB_ENV` rejects any line that is not `KEY=VALUE`.
- `skip_unless_autoembedding()` in `libs/langchain-mongodb/tests/utils.py` skips when that flag is set. Four call sites: `pytestmark` on `test_autoembedded_vectorstore.py`, `usefixtures` on `test_hybrid_retriever_autoembed`, an inline call in `test_parent_document.py`'s `embedding_param` (it branches on `request.param`, so a marker can't express it), and the same check in `langgraph-store-mongodb`.

**Absence of the flag means run.** Nothing can silently opt out of coverage, which is the bug `COMMUNITY_WITH_SEARCH` had — it was never set anywhere in CI, so its guards fired in both directions and the auto-embedding tests never ran at all.

### On the first commit

`8689396` took a different approach — probing the deployment by creating an `autoEmbed` index and reading the error. `5ca0722` replaces it, and the history is kept because "why not just detect it?" is the obvious question here.

Probing creates a real index build and drops the collection out from under it without waiting for READY. #120 documents why that is the wrong thing to spend on a yes/no question: if the Voyage key and endpoint disagree, `createSearchIndex` *succeeds* and mongot then retries a 403 forever, leaving the index in `INITIAL_SYNC`. Every test in these suites proves indexing by polling `similarity_search` until it returns `n_docs`, so a cluster in that state fails whole modules on timeouts nowhere near the cause.

It also dropped `AUTOEMBED_REQUIRED` from the first commit. That existed to turn a silent skip into a loud failure; with the default inverted to run, it is redundant.

## Test Plan

Against a local Atlas container started by the modified script:

- With the flag set, the 6 previously-failing `langchain-mongodb` tests and the 1 in `langgraph-store-mongodb` skip, each with the reason shown under `-rs`.
- With it unset, those tests genuinely attempt to run and fail against the bare container — there is no configuration in which they pass while covering nothing.
- `just integration_tests` picks the flag up via dotenv; `_test.yml:58` covers the steps that call `uv` directly.
- ruff 0.14.2 and `just typing` clean in both packages; `bash -n` clean on the script.

Not covered locally: that the tests still *pass* where auto-embedding is available. This change cannot affect that path — it adds no code that runs when the flag is unset — but a credentialed patch build is what would confirm it.

Worth noting for reviewers: a control Evergreen run on `main` reproduced the remote-Atlas failures seen on this branch, so those are independent of this PR and are being looked at separately.

### Checklist for Author

- [ ] Did you update the changelog (if necessary)? — not necessary, test-only
- [x] If there are new TODOs, has a related JIRA ticket been created? — no new TODOs

### Checklist for Reviewer


libs/langgraph-store-mongodb/tests/integration_tests/conftest.py
```diff
@@ -0,0 +1,77 @@
...
+AUTOEMBED_PROBE_IDX_NAME = "langgraph-test-index-autoembed-probe"
+
+# Substring of the server error raised when the deployment has no embedding
+# model registered.  The accompanying code is a generic ``UnknownError`` (8),
+# so the message is the only thing that distinguishes this from a real fault.
+_MODEL_NOT_REGISTERED = "not registered yet"
+
+
+@lru_cache(maxsize=1)
+def autoembedding_available() -> bool:
```
blink1073: This is a good candidate for pymongo-search-utils

blink1073 (approved):
LGTM with one comment about suggested follow up work to avoid duplication

blink1073 (approved):
LGTM

PR #460, opened 2026-09-11
fix(deepagents-vfs): strip leading slash in grep/ls/glob so agent pat…

christineastoria:
…h calls match

[JIRA TICKET INTPYTHON-1101](https://jira.mongodb.org/browse/INTPYTHON-1101)

## Summary

When MongoFilesystemBackend is used as a Deep Agent's backend, grep/ls/glob return empty even when the files exist. The agent then reports the corpus is empty.

Deep Agents' file tools pass absolute paths (with a leading /), e.g. /mongodb_vfs/journal. But SearchRouter uses the path argument to build a ^<path> regex against source_path, and stored keys are relative (no leading slash), so the regex ^/mongodb_vfs/journal matches nothing.

## Changes in this PR

Normalize the leading slash off the path prefix in search.py, consistent with what the code already does elsewhere:
- glob() already does pattern.lstrip("/")
- the S3 backend's _key() already does .lstrip("/")
- a comment in _glob already notes "Leading '/' is not meaningful for S3 keys, which are always relative"

Changes (all in SearchRouter):
- ls(): path.rstrip("/") → path.strip("/")
- _glob(): path.rstrip("/") → path.strip("/")
- grep(): path = path.lstrip("/") before it's used as a ^<path> regex prefix

Pure path normalization — no change to inputs, credentials, or query surface; the regex is still built with re.escape on the normalized value.

## Test Plan

Added TestSearchRouterLeadingSlash (unit, mongomock — no Atlas) asserting ls/glob/grep return the same non-empty results for /docs as for docs. These fail before the fix (absolute path → empty) and pass after.

uv run pytest tests/unit_tests/test_search_router.py -q
19 passed

### Checklist for Author

- [x] Has a MongoDB Employee run [the patch build of this PR](https://github.com/mongodb-labs/ai-ml-pipeline-testing?tab=readme-ov-file#running-a-patch-build-of-a-given-pr)?

### Checklist for Reviewer


caseyclements:
[Passing patch build](https://spruce.corp.mongodb.com/version/6aa4640c13b42d00077e6cd6/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

caseyclements:
@anujbpanchal As the original contributor, would you mind taking a look at this? Please add comments, request changes, or LGTM, and I'll review Tuesday.

caseyclements:
@christineastoria Thank you for submitting this PR! Would you please fix the lint/format errors?
You can run `just lint` and ruff will auto-format for you. Then you just need to add the few changes, commit, and push.

anujbpanchal:
> Anuj Panchal (@anujbpanchal) As the original contributor, would you mind taking a look at this? Please add comments, request changes, or LGTM, and I'll review Tuesday.

@caseyclements LGTM

PR #461, opened 2026-09-12
INTPYTHON-1103 Wait for the fulltext index in integration tests

caseyclements:
[INTPYTHON-1103](https://jira.mongodb.org/browse/INTPYTHON-1103)

## Summary

Integration tests query Atlas Search fulltext indexes before mongot has indexed the documents the fixtures just inserted. The tests then see an empty result set and fail.

The `collection` fixtures create their indexes under a guard:

```python
if not any(SEARCH_INDEX_NAME == ix["name"] for ix in clxn.list_search_indexes()):
    create_fulltext_search_index(..., wait_until_complete=TIMEOUT)
```

Search indexes survive between runs, so the guard is false on every run after the first and the only `wait_until_complete` call is skipped. Even on the first run that call waits for index *creation*, not for documents inserted afterwards.

The `vectorstore` fixtures then insert documents. `PatchedMongoDBAtlasVectorSearch` polls the **vector** index only. Nothing waits for the fulltext index.

`test_hybrid_search_weighted_rrf` fails deterministically on `assert single_text_score > 0`. The other hybrid tests lose their fulltext leg, degrade to vector-only ranking, and fail their ordering assertions. `test_hybrid_retriever_nested` was masked by `@flaky(max_runs=5, min_passes=4)`.

This PR also fixes a second, unrelated failure in the same file. `pymongo_search_utils.drop_vector_search_index` waits on this predicate:

```python
predicate=lambda: len(list(collection.list_search_indexes())) == 0
```

That is "the collection has zero search indexes", not "this index is gone". `test_hybrid_retriever_auto_create_index` drops `text_index` on a collection that keeps `vector_index`, so the predicate can never hold. The test waits the full 60s and fails, even though the drop succeeded. Because the drop *did* take effect, the next run skips the drop and passes, so the test alternates between passing and failing.

These failures force full-suite retries in `execute-tests.sh` (`MAX_ATTEMPTS=3`), which contributes to `test-langchain-python-remote` exceeding `exec_timeout_secs: 3600`.

## Changes in this PR

**`tests/utils.py`** — two new helpers, both additive. No existing function changed.

- `wait_for_fulltext_index(collection, index_name, path, timeout=TIMEOUT)` polls a `$search` using the `exists` operator with `$count` until the index reports as many documents as the collection holds. `exists` matches every indexed document that has the field, so the wait does not depend on the corpus wording.
- `drop_search_index_and_wait(collection, index_name, timeout=TIMEOUT)` drops one index and waits for that index name to disappear.

**`tests/integration_tests/test_rerank.py`**

- New module-scoped `fulltext_indexed` fixture, which depends on `vectorstore` so the insert has already happened.
- Added it to `test_fulltext_retriever_rerank` and `test_hybrid_retriever_rerank`, the only two tests there that use the fulltext index.

**`tests/integration_tests/test_retrievers.py`**

- Three fixtures, one per collection and index pair: `fulltext_indexed`, `fulltext_indexed_autoembed`, and `fulltext_indexed_nested`.
- Added them to `test_hybrid_retriever`, `test_hybrid_retriever_autoembed`, `test_hybrid_retriever_deprecated_top_k`, `test_hybrid_retriever_nested`, `test_hybrid_search_weighted_rrf`, and `test_fulltext_retriever`.
- Replaced the inline wait loop in `test_fulltext_retriever` with the fixture. That loop protected only its own test, and it ran after every hybrid test in the file.
- Both `*_auto_create_index` tests now call `drop_search_index_and_wait`.

`PatchedMongoDBAtlasVectorSearch` is unchanged, and so is the existing vector wait.

## Test Plan

These are the tests. The change adds no new production code.

- `test_rerank.py` passed twice in a row against a remote Atlas cluster.
- `test_retrievers.py` passed. Please run it **twice in a row** when reviewing: the drop bug only appears on a run that finds both `text_index` and `vector_index` present, so a single green run does not exercise it.
- The waits cost nothing once the index has caught up. The first poll returns and the fixture exits. The drop change removes a guaranteed 60s timeout rather than adding time.

### Checklist for Author

- [x] If there are new TODOs, has a related JIRA ticket been created?

### Checklist for Reviewer


caseyclements:
[First passing build](https://spruce.corp.mongodb.com/version/6aa4a79857978900076e95ba/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

[Second passing build](https://spruce.corp.mongodb.com/version/6aa4b0e26b3bf50007cb7eab/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

caseyclements:
Filed [PYTHON-6095](https://jira.mongodb.org/browse/PYTHON-6095) for the underlying bug in `pymongo_search_utils.drop_vector_search_index`.

Its wait predicate is `len(list(collection.list_search_indexes())) == 0`, which means "the collection has zero search indexes" rather than "this index is gone". Any caller that drops one index from a collection that keeps another waits the full timeout and then raises, even though the drop succeeded.

The `drop_search_index_and_wait` helper added here is a test-side workaround. Delete it once PYTHON-6095 is fixed.

libs/langchain-mongodb/tests/utils.py
```diff
@@ -49,6 +49,54 @@ def create_database() -> MongoDBDatabase:
     return MongoDBDatabase(client, DB_NAME)
 
 
+def wait_for_fulltext_index(
```
NoahStapp, line 52: This doesn't look langchain-specific, should we move this into `pymongo-search-utils` instead?

NoahStapp (changes requested):

libs/langchain-mongodb/tests/utils.py
caseyclements: > This doesn't look langchain-specific, should we move this into `pymongo-search-utils` instead?

[PYTHON-6095](https://jira.mongodb.org/browse/PYTHON-6095). Yes, and this ticket was created to track that. My job here is to get all the tests green.

NoahStapp: That ticket doesn't mention `wait_for_fulltext_index`, only the temporary `drop_search_index_and_wait` helper that doesn't use `wait_for_fulltext_index` at all.

Regarding [PYTHON-6095](https://jira.mongodb.org/browse/PYTHON-6095), why don't we do the fix in `pymongo-search-utils` now and release a new version of it instead of a temporary fix here that we'll have to revert once that broader fix is implemented?

caseyclements: > That ticket doesn't mention `wait_for_fulltext_index`, only the temporary `drop_search_index_and_wait` helper that doesn't use `wait_for_fulltext_index` at all.
> 
> Regarding [PYTHON-6095](https://jira.mongodb.org/browse/PYTHON-6095), why don't we do the fix in `pymongo-search-utils` now and release a new version of it instead of a temporary fix here that we'll have to revert once that broader fix is implemented?

Oh. My bad. I'm under the weather today. The reason that I did that was because my work doing ABX Greener Build was timeboxed until today. I don't have capacity for a pymongo-search-utils release.

caseyclements:
[PYTHON-6095](https://jira.mongodb.org/browse/PYTHON-6095) now has an open [PR in pymongo-search-utils](https://github.com/mongodb-labs/pymongo-search-utils/pull/47). 

@NoahStapp  I will refactor this PR to point to it.

